PR #61, opened 2015-04-09
capture git versioning information in manifest

mvp:
For every generated jar, capture git versioning information in Java manifest.
This allows to tell at runtime exact git commit id that this jar was built from.


martint:
Merged

PR #71, opened 2015-08-26
Update to joda-time 2.8.2

wrmsr:
Per https://stackoverflow.com/questions/32058431/aws-java-sdk-aws-authentication-requires-a-valid-date-or-x-amz-date-header joda-time < 2.8.1 is broken on 8u60 which breaks aws connectivity in presto. Per http://www.joda.org/joda-time/upgradeto282.html the only other change is newer DateTimeZone data.


martint:
merged, thanks!

PR #75, opened 2015-10-15
testng - exclude guice dependency

kokosing:
testng - exclude guice dependency

testng@6.9.6 is depending on guice@4.0-no_aop. It is common that this
dependency (guice) is conflicting as many projects is using guice with
AOP support.


kokosing:
@electrum @martint Is there any chance to get this in?

martint:
Merged, thanks!

PR #78, opened 2016-01-24
Add dependencies for AssertJ and Truth

electrum:


martint:
Why Truth? As far as I can tell it has no significant advantage over assert4j

electrum:
Seems like we might as well add both, making it easier to try out in different projects.

martint:
I don't think we should be adding libraries to airbase until we decide to go with them "officially". If projects want to try them, they can just do it.

PR #93, opened 2016-10-01
Upgrade JMH to 1.15

monkey-mas:
It'd be nice to use the latest version as of today.

ref. http://mail.openjdk.java.net/pipermail/jmh-dev/2016-September/002371.html


martint:
Merged, thanks!

PR #94, opened 2016-11-02
Update Surefire to 2.19.1

electrum:


pom.xml
```diff
@@ -465,7 +464,7 @@
                         <threadCount>${air.test.thread-count}</threadCount>
                         <argLine>
                             ${argLine}
-                            -Duser.timezone=${air.test.timezone}
```
dain: Did you mean to remove this?

electrum: Yes. IntelliJ needs it in system properties to pass it as an argument, and Surefire warns if you have it both places. I tested this and it seems to work fine. I'm actually not sure why we added it here in the first place.

PR #97, opened 2016-11-09
Add maven wrapper

kokosing:
Add maven wrapper


electrum:
Why do we need this? Also, we don't need the smart builder or huge heap sizes, since there's nothing to build aside from the POM.

kokosing:
I don't know if we need this. I just wanted to `mvn install` this locally and I don't have maven installed system-wide. Generally it is not a big issue to not have a maven wrapper for me, but it is just cumbersome that for some project we do use it and for some not.

kokosing:
@electrum comments addressed

hgschmie:
that is the pattern that gradle established where every project drags around its own copy of all the build tools and splats them all over the disk to avoid telling the user "you need to install maven if you want to build this software" (iaw run a `brew install maven`. 

personally I see next to no value having every single repo on my disk trying to download another copy of maven 3.3.9 to install.

electrum:
We use Maven Wrapper for Presto because we need the latest version, newer than what Travis CI provides (currently 3.2.5). For this project, the minimum version is 3.2.3, which was released over two years ago.

As such, we'd rather not add non-required cruft here (it's many times larger than the rest of the project!)

kokosing:
Sounds good. Thanks for explanation.

PR #110, opened 2017-08-11
Add JUnit 5 support

ebd2:
With this change, we can now run both JUnit-5 and TestNG tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
-Airbase 67
+Airbase 68
 
+* Add support for JUnit-5
+
+Airbase 67
 * Ban `com.google.code.findbugs:annotations` dependency
 * Add dependencies:
   - FindBugs / SpotBugs annotations 3.1.0-RC5
--- a/pom.xml
+++ b/pom.xml
@@ -183,6 +183,10 @@
         <dep.objenesis.version>1.3</dep.objenesis.version>
         <dep.slice.version>0.10</dep.slice.version>
         <dep.jmh.version>1.15</dep.jmh.version>
+        <dep.junit.jupiter.version>5.0.0-RC2</dep.junit.jupiter.version>
+        <dep.junit.jupiter.platform.version>1.0.0-RC2</dep.junit.jupiter.platform.version>
+        <dep.junit.version>4.12</dep.junit.version>
+        <dep.junit.vintage.version>${dep.junit.version}.0-RC2</dep.junit.vintage.version>
 
         <!-- license headers -->
         <air.license.owner>${project.organization.name}</air.license.owner>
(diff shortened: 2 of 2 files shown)


ebd2:
@electrum

pom.xml
```diff
@@ -183,6 +183,10 @@
         <dep.objenesis.version>1.3</dep.objenesis.version>
         <dep.slice.version>0.10</dep.slice.version>
         <dep.jmh.version>1.15</dep.jmh.version>
+        <dep.junit.jupiter.version>5.0.0-RC2</dep.junit.jupiter.version>
```
electrum: I think our convention is to name these `dep.NAME.version` so let's call this `dep.junit-jupiter.version`. That will match with `assertj-guava`, etc.

pom.xml
```diff
@@ -506,6 +510,23 @@
                             -XX:+HeapDumpOnOutOfMemoryError
                         </argLine>
                     </configuration>
+					<dependencies>
```
electrum: The indentation is off. Maybe this is using tabs?

pom.xml
```diff
@@ -183,6 +183,10 @@
         <dep.objenesis.version>1.3</dep.objenesis.version>
         <dep.slice.version>0.10</dep.slice.version>
         <dep.jmh.version>1.15</dep.jmh.version>
+        <dep.junit.jupiter.version>5.0.0-RC2</dep.junit.jupiter.version>
+        <dep.junit.jupiter.platform.version>1.0.0-RC2</dep.junit.jupiter.platform.version>
```
electrum: Let's call this `junit-platform` since that matches their Maven artifacts (which don't contain the word "jupiter").

pom.xml
```diff
@@ -183,6 +183,10 @@
...
+        <dep.junit.version>4.12</dep.junit.version>
```
electrum: This is unused. Did you intend to add a dependency for this?

pom.xml
```diff
@@ -183,6 +183,10 @@
...
+        <dep.junit.vintage.version>${dep.junit.version}.0-RC2</dep.junit.vintage.version>
```
electrum: This is also unused. I'm not sure how it's intended to be used, but I'd copy the version rather than appending what appears to be a very specific suffix that will likely change in the future (when it's no longer RC).

pom.xml
ebd2: This is solely there to provide the JUnit version for the following line. See the comment on that one.

pom.xml
ebd2: The JUnit vintage stuff lets JUnit5 run tests from prior versions of JUnit. I'm fine with dropping it entirely since it's unlikely that we're ever going to want to introduce JUnit tests for older versions of JUnit into the Presto codebase.

I took the version number stuff from
https://github.com/junit-team/junit5-samples/blob/master/junit5-maven-consumer/pom.xml#L12

ebd2:
@electrum: Feedback addressed, back to you.

CHANGES.md
```diff
@@ -1,5 +1,8 @@
-Airbase 67
+Airbase 68
+
+* Add support for JUnit-5
```
electrum: Nit: "JUnit 5"

electrum:
Merged, thanks!

PR #121, opened 2017-10-25
Upgrade Logback to v1.2.3

highker:
The latest Logback contains fixes to reduce the lock contention
while logging events to a file.


dain:
Merged, thanks!

PR #129, opened 2017-12-17
Prevent Surefire from adding java.se.ee module on Java 9

electrum:


pom.xml
```diff
@@ -502,6 +502,7 @@
                             -Xms${air.test.jvmsize}
                             -XX:OnOutOfMemoryError="kill -9 %p"
                             -XX:+HeapDumpOnOutOfMemoryError
+                            --add-modules java.base
```
martint: Is this in a java 9-only profile? Otherwise, it will fail with Java 8, no?

electrum: Good catch, let me figure out how to do that.

electrum:
Updated. Also added a commit to remove the m2e settings, which is not included in basepom.

PR #130, opened 2017-12-21
Update jmh to 1.19

martint:
It fixes a bug when processing the output of perfasm in Java 9


electrum:
Fixed in 43e324f5c6d460f5333ffed50d9c0da5235f282d

PR #133, opened 2018-01-16
Add checkstyle for empty line before {

kokosing:


kokosing:
See https://github.com/prestodb/presto/pull/9701/commits/2666efeb255e8048a82dce597f3d5a8035c5224e

kokosing:
See https://github.com/prestodb/presto/pull/9755.

Nice thing about this is that it removes things like (main goal of this PR and the most common misformatting): 
```
public type method(args)

{
  //code
}
```

But also it catches things like custom scopes in method or initialization blocks, which are rarely used in Presto codebase so maybe it would be nice to avoid weird syntax, see:
```
public class Class {
    private final Type type;

    {
        // code
       type = ...;
    }

    public void m() {
         {
               // code
         }

         {
              // code
         }
    }
}
```

kokosing:
CC: @electrum

martint:
Can we catch the misformatting problem without limiting the use of custom scopes? I don’t know about others, but use those a lot while prototyping, developing and refactoring code, so excluding them would be a step backward.

martint:
Also, instance initializers are necessary for anonymous inner classes, which cannot have constructors.

kokosing:
@martint I do not think any of these is possible to be achieved in reliable way. This check style is using simple regular expression. We can either accept all of it of none. Things mentioned by you would be still possible, but would look like:

```
public class Class {
    private final Type type;
    {
        // code
       type = ...;
    }

    public void m() {
         {
               // code
         }
         {
              // code
         }
    }
}
```

While custom scopes look OK, while instance initializers may look like a method (`s/;/()/`).

electrum:
Most of these regex checks would be better done as real checkers. We could write a plugin with these checks and try to get them contributed upstream.

PR #135, opened 2018-01-24
Forbid import of format methods other than String.format

findepi:
This is to prevent cases like https://github.com/prestodb/presto/pull/9807

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -49,6 +49,10 @@
         <property name="format" value="^import static .*\.(of|copyOf|valueOf);$" />
         <property name="message" value="The following methods may not be statically imported: of, copyOf, valueOf" />
     </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="^import static (?!java\.lang\.String\.format;).*\.format;" />
+        <property name="message" value="Only 'format' from java.lang.String may be statically imported" />
+    </module>
 
     <module name="RegexpSingleline">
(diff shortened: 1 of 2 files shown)


findepi:
cc @electrum

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -49,6 +49,10 @@
         <property name="format" value="^import static .*\.(of|copyOf|valueOf);$" />
         <property name="message" value="The following methods may not be statically imported: of, copyOf, valueOf" />
     </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="^import static (?!java\.lang\.String\.format;).*\.format;" />
```
electrum: Should the pattern end in `$`? We do that for the others

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -49,6 +49,10 @@
...
+        <property name="message" value="The format method cannot be statically imported" />
```
electrum: How about making the message
```
Only 'format' from java.lang.String may be statically imported
```
This should give the user a better hint about what is wrong

electrum:
Can you update CHANGES? Let's add a new type of update:
```
* Checkstyle updates:
  - Forbid import of format methods other than String.format

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
findepi: Why would i want it to end with `$`?

findepi:
applied

findepi:
Ping

electrum:
Merged, thanks!

electrum:
I had never seen `?!`. That's really cool. That can probably be used to clean up several other checks.

PR #138, opened 2018-02-27
Update Guava and Guice

electrum:


airbase/pom.xml
```diff
@@ -166,7 +166,7 @@
         <dep.packaging.version>0.163</dep.packaging.version>
 
         <!-- Dependency versions that should be the same everywhere. -->
-        <dep.guice.version>4.0</dep.guice.version>
+        <dep.guice.version>4.1.0</dep.guice.version>
```
martint: Guice 4.2 is supposed to come out over the next few days.

electrum: I'll update to that and put up a separate PR for just Guice

PR #143, opened 2018-06-03
Require simplified usages of annotations

electrum:


martint:
What’s a “simplified usage”?

electrum:
It disallows unnecessary array brackets or parenthesis. Unfortunately, it doesn’t detect unnecessary values=. See http://checkstyle.sourceforge.net/config_annotation.html#AnnotationUseStyle

PR #149, opened 2018-09-04
Upgrade jackson to 2.8.10

zhenxiao:


zhenxiao:
@kokosing this is to upgrade jackson to 2.8.6, so that elasticsearch connector and presto main jackson version aligned

airbase/pom.xml
```diff
@@ -173,7 +173,7 @@
         <dep.javax-inject.version>1</dep.javax-inject.version>
```
kokosing: Please update `CHANGES.md` file

zhenxiao: sure, get it updated

electrum:
We should probably update to the latest 2.8 release, 2.8.9

Can you put up test-only PRs for Presto and Airlift to update the Jackson version there (add it to the `dependencyManagement` of the root POM) so that we can verify that Travis passes? Then we can update it here. If there are problems that will require work to fix, we want to have that work ready to go before updating Airbase (otherwise it would block any changes in Airbase).

zhenxiao:
thank you, @electrum 
get it updated to 2.8.10
here are the 2 testing PRs:
https://github.com/airlift/airlift/pull/670
https://github.com/prestodb/presto/pull/11455

electrum:
Thanks! We are now on the latest, 2.9.7

PR #150, opened 2018-09-10
Add checkstyle check to enforce whitespace before inline array

tdcmeehan:
initialization

Just trying to quickly automate some code review feedback I received.  If approved I'll also open a small corresponding PR in Presto to fixup the existing usages.


electrum:
Merged, thanks!

PR #151, opened 2018-09-17
Require release dependencies

arhimondr:
Add maven-enforcer-plugin rule that checks the dependencies
and fails if any snapshots are found.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 84
+
+* Forbid snapshot versions for dependencies
+
 Airbase 83
 
 * Dependency updates:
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+Airbase 84
+
+* Add rule that checks the dependencies and fails if any snapshots are found
```
electrum: How about
```
* Disallow snapshot versions for dependencies

arhimondr:
Example error message: https://gist.github.com/arhimondr/544683648eb4edacafd0ee68fceba410

arhimondr:
Yes, it complicates somehow. But it protects you from using a SNAPSHOT dependencies by mistake.

PR #152, opened 2018-10-04
Forbid static import of Optional's members

findepi:


findepi:
Presto code base will need to be updated as well

PR #162, opened 2018-10-29
Add airlift to dependency management

findepi:


dain:
@findepi IIRC we used to do this, removed the code because it creates a circular dependency between the projects that was too hard to maintain. I'm hoping @electrum or @martint remember the details.

findepi:
I didn't realize this won't work.
Is it "logical circular dependency" or something more of a technical level?

electrum:
I'm actually not sure if we ever tried this. I think it can work, since we have Slice in Airbase and Slice uses Airbase. Can you try building Airlift with this? BTW, does it cause any problems to not define the version in Airbase?

I think we avoided it because it creates a logical circular dependency, but it seems fine to me now.

PR #163, opened 2018-11-01
Generate parameters names reliably

findepi:
Due to https://youtrack.jetbrains.com/issue/IDEA-194601, when using
```
<artifactId>maven-compiler-plugin</artifactId>
<configuration>
    <parameters>true</parameters>
...
```

parameters are not generated when code is compiled in IntelliJ.

Also, manual IDE configuration is not an option, since the configuration
gets reset each time IDE auto-loads pom changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 87
+
+* Generate parameters names also when compiling in IntelliJ
+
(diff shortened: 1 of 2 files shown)


electrum:
Can you update CHANGES?

PR #164, opened 2018-11-06
Make possible to overwrite readme file location and type

kokosing:
Make possible to overwrite readme file location and type

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -205,6 +205,10 @@
 
         <!-- Checkstyle -->
         <air.checkstyle.config-file>checkstyle/airbase-checks.xml</air.checkstyle.config-file>
+
+        <!-- Readme for launcher -->
+        <air.readme.file>README.txt</air.readme.file>
+        <air.readme.type>txt</air.readme.type>
     </properties>
 
     <build>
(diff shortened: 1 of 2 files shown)


airbase/pom.xml
```diff
@@ -197,6 +197,10 @@
 
         <!-- Checkstyle -->
         <air.checkstyle.config-file>checkstyle/airbase-checks.xml</air.checkstyle.config-file>
+
+        <!-- Readme -->
```
electrum: We should be more specific:
```suggestion
        <!-- Readme for launcher -->
```

electrum:
What's the motivation for this change? I think there was a specification for the tarball format.

cc @martint

kokosing:
> What's the motivation for this change?

We use airbase in one of projects and we had dummy README.txt (and proper README.md) just to supply this check this. Not a big deal, but it would be nice to attach proper readme file.

airbase/pom.xml
```diff
@@ -198,7 +198,7 @@
         <!-- Checkstyle -->
         <air.checkstyle.config-file>checkstyle/airbase-checks.xml</air.checkstyle.config-file>
 
-        <!-- Readme -->
+        <!-- Readme for launcher -->
```
kokosing: This is going to be squashed.

kokosing:
Updated

electrum:
Merged, thanks!

PR #167, opened 2018-11-14
Upgrade to bval 2.0.0

findepi:


findepi:
@electrum Validation 2.0 brings ability to validate List and Optional elements, for example `List<@Pattern(...) String>`.

airbase/pom.xml
```diff
@@ -169,8 +169,9 @@
         <dep.slf4j.version>1.7.25</dep.slf4j.version>
         <dep.logback.version>1.2.3</dep.logback.version>
         <dep.javax-inject.version>1</dep.javax-inject.version>
-        <dep.javax-validation.version>1.1.0.Final</dep.javax-validation.version>
+        <dep.javax-validation.version>2.0.1.Final</dep.javax-validation.version>
```
electrum: Add this to CHANGES

airbase/pom.xml
```diff
@@ -1126,18 +1127,16 @@
             </dependency>
 
             <!-- misc stuff -->
-
-            <!-- bval 1.1.1 depends on commons-lang3 3.3.2, which is incompatible with Java 9 -->
             <dependency>
```
electrum: We can delete this dependency now. Based on the comment, it was added to force a newer version (than bval used) that is compatible with Java 9.

airbase/pom.xml
```diff
@@ -1126,18 +1127,16 @@
...
-                <version>3.4</version>
+                <version>3.7</version>
             </dependency>
 
             <dependency>
                 <groupId>org.apache.bval</groupId>
                 <artifactId>bval-jsr</artifactId>
-                <version>1.1.1</version>
+                <version>${dep.bval.version}</version>
                 <exclusions>
```
electrum: None of these exclusions should be needed now. I added it to a new project and `dependency:tree` shows no dependencies. Although it's strange that `commons-weaver-privilizer-api` is declared as a dependency but Maven doesn't pick it up.

electrum: Ah, I see it's not picked up because the parent POM declares it as `provided` scope in `dependencyManagement`. I didn't realize you could do that (and it seems strange).

electrum:
Thinking about this more, since it appears the API didn't change, the only requirement is updating in Airbase (or otherwise bumping the dependency version). We don't need to change Airlift at all. So maybe leaving it in Airbase is better?

findepi:
There is additional bonus when this stays in airbase -- when i want to do validation outside of airlift `configuration`, i will pull the matching bval version easily. And my use-case just happens to be that.

PR #170, opened 2018-12-04
Checkstyle updates

electrum:


airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -124,6 +124,12 @@
         <module name="RightCurly">
             <property name="option" value="alone" />
         </module>
+        <module name="EmptyLineSeparator">
+            <property name="allowNoEmptyLineBetweenFields" value="true" />
+            <property name="tokens" value="
+                IMPORT, STATIC_IMPORT, CLASS_DEF, INTERFACE_DEF, ENUM_DEF,
+                STATIC_INIT, INSTANCE_INIT, METHOD_DEF, CTOR_DEF, VARIABLE_DEF" />
```
martint: What's the effect of `VARIABLE_DEF` here? Will it force a new line between all variable definitions?

findepi: same for  IMPORT and STATIC_IMPORT

electrum: It seems to just be for fields. I ran it on the Airlift and Presto code bases and the checks passed. It also caught the things I expected it to catch.

This is the default list expect for “package”, which we can’t use since we don’t put a blank line after the copyright header.

PR #178, opened 2019-02-01
Update maven-source-plugin to 3.0.1

findepi:
This fixes skipping sources with `-Dmaven.source.skip=true`. The flag is
said documented as supported "since 2.2", but was not effective in 2.2.1.


findepi:
cc @electrum

PR #181, opened 2019-02-18
Apply timezone/region settings reliably

findepi:
Previously, when running tests in Presto (`./mvnw test -pl presto-main`)
on my machine, `user.timezone` would be set to my local time zone
and `user.region` would not be set at all.

When running tests in IntelliJ (my version), both versions (before the
change and after) seem to work the same.

cc @electrum @sopel39

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -415,9 +415,6 @@
                     <configuration>
                         <systemPropertyVariables>
                             <sun.jnu.encoding>${project.build.sourceEncoding}</sun.jnu.encoding>
-                            <user.timezone>${air.test.timezone}</user.timezone>
-                            <user.language>${air.test.language}</user.language>
-                            <user.region>${air.test.region}</user.region>
                             <java.awt.headless>true</java.awt.headless>
                             <java.util.logging.SimpleFormatter.format>%1$tY-%1$tm-%1$tdT%1$tH:%1$tM:%1$tS.%1$tL%1$tz %4$s %5$s%6$s%n</java.util.logging.SimpleFormatter.format>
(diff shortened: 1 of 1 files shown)


airbase/pom.xml
```diff
@@ -433,6 +430,9 @@
                             -XX:+ExitOnOutOfMemoryError
                             -XX:+HeapDumpOnOutOfMemoryError
                             -XX:-OmitStackTraceInFastThrow
+                            -Duser.timezone=${air.test.timezone}
```
sopel39: I wonder if other properties get's propagated

findepi:
@martint

airbase/pom.xml
martint: Indeed. Do we need to do the same for java.awt.headless and the java.util.logging.SimpleFormatter property?

findepi: If i understand this correctly, the difference between `argLine` and `systemPropertyVariables` is that the latter gets applied by some Maven class, from within forked JVM, so after some Java code has already run. This is probably what made setting timezone ineffective -- it's too late.
It might be the case that ` java.awt.headless` doesn't matter until you try to use some graphics classes, but I agree it looks safer to just use `argLine` exclusively.

findepi: I realized I am wrong. This has nothing to do with `argLine` vs `systemPropertyVariables` -- just presto-main defines `systemPropertyVariables` and overrides `systemPropertyVariables` defined in airbase. Instead of merging into it. I'll try to fix this in Presto instead

findepi:
Superseded by https://github.com/prestosql/presto/pull/528

PR #182, opened 2019-03-05
Require package names not to contain uppercase letters nor underscores

findepi:


findepi:
Presto code base needs to be updated.

findepi:
@electrum @martint thoughts?

martint:
Sounds reasonable to me.

PR #183, opened 2019-06-02
Alllow multi character type variable names

dain:


airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -165,10 +165,10 @@
         <module name="ParameterName" />
         <module name="StaticVariableName" />
         <module name="ClassTypeParameterName">
-            <property name="format" value="^[A-Z][0-9]?$" />
+            <property name="format" value="^[A-Z][A-Z0-9]*$" />
         </module>
         <module name="MethodTypeParameterName">
-            <property name="format" value="^[A-Z][0-9]?$" />
+            <property name="format" value="^[A-Z][A-Z0-9]*$" />
```
findepi: In the past I found it convenient to name my params camel case rather than all-caps (in the case where single-char param names were not enough).
Since we now are moving from single-char (plus optional number, whatever) to multi-char, we can make any choice we want.

findepi:
typo in cmt msg "Alllow"

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
dain: I think it might be confusing to mix the type variable and normal class name namespace.  I could imagine wanting an underscore in some cases.... I suggest that we expand this more when we find another use case.

BTW my use case was for a between function, I wanted to use `<V, MIN, MAX>`

martint: > In the past I found it convenient to name my params camel case rather than all-caps (in the case where single-char param names were not enough).

Yeah, I like doing that, too. For example, I once was working on a library to manipulate graphs, and I used NodeType and EdgeType as my type variables.

martint: > I think it might be confusing to mix the type variable and normal class name namespace.

It *is* the same namespace, though. Type variables have precedence over class names, just like local variables have precedence over instance variables. IDEs like IntelliJ will highlight them differently.

findepi: So it's 2-1 :)

dain: @martint yes, it _is_ the same namespace, which is why it can be confusing since you can shadow types.  Anyway, I don't feel strongly.  I just have a preference for relaxing when we need stuff.

electrum: The IDE does highlight generic parameters differently, though that doesn't help on GitHub, etc. The advantage of all uppercase is that you know visually it can't be a class name (at least one that follows our naming convention).

martint: > The IDE does highlight generic parameters differently, though that doesn't help on GitHub, etc. The advantage of all uppercase is that you know visually it can't be a class name (at least one that follows our naming convention).

But that's the same argument we've made for ages against having a naming convention for local variables vs instance methods like many projects do (not that I'm advocating for that!). It's easy to spot them in an IDE, but not in github, etc.

PR #186, opened 2019-09-28
Bump Jackson to 2.9.10

Fokko:
Dear maintainers. 

I'd like to bump Jackson to 2.9.9 and databind to 2.9.9.3 due to discovered CVE's in databind: https://www.cvedetails.com/vulnerability-list.php?vendor_id=15866&product_id=42991&version_id=&page=1&hasexp=0&opdos=0&opec=0&opov=0&opcsrf=0&opgpriv=0&opsqli=0&opxss=0&opdirt=0&opmemc=0&ophttprs=0&opbyp=0&opfileinc=0&opginf=0&cvssscoremin=0&cvssscoremax=0&year=0&month=0&cweid=0&order=1&trc=23&sha=5425c8855ccdf6005f8b83f6d90fed738cf8a8e8

This triggered my vulnerability scanner, and it would be great to get this patched. Bumping to 2.10 isn't possible yet since we're supporting Java8. Would it be possible to create a new release as well? So I can downstream this in other projects. Please let me know.

Cheers, Fokko

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+Airbase 92
+
+* Dependency updates:
+  - Jackson 2.9.10 (from 2.9.8)
+
 Airbase 91
 
 * Require package names not to contain uppercase letters nor underscores.
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -181,7 +181,7 @@
         <dep.javax-validation.version>2.0.1.Final</dep.javax-validation.version>
         <dep.javax-servlet.version>3.1.0</dep.javax-servlet.version>
         <dep.bval.version>2.0.0</dep.bval.version>
-        <dep.jackson.version>2.9.8</dep.jackson.version>
+        <dep.jackson.version>2.9.10</dep.jackson.version>
         <dep.jmxutils.version>1.19</dep.jmxutils.version>
         <dep.cglib.version>3.2.5</dep.cglib.version>
         <dep.joda.version>2.9.9</dep.joda.version>


findepi:
Thanks @Fokko for bringing attention to this issue.
i am aware jackon-core 2.9.9.3 wasn't released and we cannot simply change `dep.jackson.version` to 2.9.9.3.
At the same time, introducing new `dep.` property means trouble for downstream projects, e.g. ones that define `dep.jackson.version=2.10...`
Let's have discussion about alternatives.

> Would it be possible to create a new release as well?

It sounds like you need this urgently.
Would it be a short term fix for your problem if you define `dep.jackson.version=2.9.9` in your project and override databind version to 2.9.9.3?

Fokko:
Thank you @findepi for the prompt reply. I fully agree with you, having another dep might introduce downstream problems. I'm trying to bump Apache Avro to 1.9.1 in Presto, which relies on the 2.9 branch of Jackson. In that case I would suggest to bump to Jackson 2.9.10 right away. WDYT?

findepi:
@Fokko Did you try building Airlift and Presto with this airbase change? 
In particular, did it pass dependency checks?

Fokko:
Yes, I've just ran the tests, and it did pass. I've installed `airbase` locally using `mvn clean install`. And then updated Presto to use `92-SNAPSHOT` as a parent. I've ran `mvn clean install -DskipTests`, it compiled and ran the Maven enforce checks.

I've also noticed that Jackson dependency was marked as ignored: https://github.com/prestosql/presto/blob/master/pom.xml#L1285

After removing this, and installing again, it also ran fine.

When I build from the current `91` parent, and I remove the ignore line mentioned above, then it fails. So after this, we can also get rid of the Jackson exclusion.

Fokko:
@findepi Any more concerns? It looks like the CI has some weird error and it looks related to Travis.

findepi:
Re: travis -- 
first failure is easy (https://github.com/airlift/airbase/pull/187 fixes it)
the other two - i retriggered but it failed too.

cc @electrum

Fokko:
The CI is still sad. I've opened https://github.com/airlift/airbase/pull/188 to fix this.

Fokko:
Rebased onto master

electrum:
The final change looks good. Please squash the changes, then I can merge this.

Fokko:
@electrum I've squashed the commits.

PR #187, opened 2019-10-01
Replace Oracle Java 8 with OpenJDK

findepi:
Oracle Java 8 is no longer available on Travis.


findepi:
This fixes Java 8 configuration.

`oraclejdk9` and `openjdk-ea` are still failing.
I think we should drop 9.

PR #193, opened 2019-11-06
Updated git commit id plugin

mosabua:
- to version 3.0.1
- requires Java 8 and Maven 3.3.9+
- allows skipping with -Dmaven.gitcommitid.skip=true
- results in faster build since remote git fetch is avoided
- see https://github.com/git-commit-id/maven-git-commit-id-plugin/blob/master/maven/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L241
- https://github.com/git-commit-id/maven-git-commit-id-plugin

Signed-off-by: Manfred Moser <mmoser@apache.org>
Signed-off-by: Manfred Moser <manfred@simpligility.com>

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -513,7 +513,7 @@
                 <plugin>
                     <groupId>pl.project13.maven</groupId>
                     <artifactId>git-commit-id-plugin</artifactId>
-                    <version>2.1.13</version>
+                    <version>3.0.1</version>
                     <configuration>
                         <dateFormat>yyyy-MM-dd'T'HH:mm:ssZZ</dateFormat>
                         <gitDescribe>


mosabua:
Note that updating airbase in presto for example to get this change also pulls in checkstyle updates, which result in errors there at the moment.

findepi:
> Note that updating airbase in presto for example to get this change also pulls in checkstyle updates, which result in errors there at the moment.

https://github.com/airlift/airbase/pull/194

electrum:
Thanks. Please also update the `CHANGES` file in a *Plugin updates* section.

For the commit message, use the imperative mood, so say *Update* rather than *Updated*. We follow these guidelines: https://chris.beams.io/posts/git-commit/

electrum:
Put the version in the commit title rather than in the description, since that's the most relevant part of the commit and thus can be found quickly:

> Update Git commit ID plugin to 3.0.1

We don't need to document the Java version since Airbase already requires Java 8 for running Maven. Before this commit, add a pre commit that updates `air.maven.version` to `3.3.9` and also documents this in `CHANGES`:

> Set minimum Maven version to 3.3.9

Then we don't need to document the Maven requirement for this plugin.

(We need to improve the CI coverage for this project so that we test with both the minimum Maven version and with newer versions.)

electrum:
I updated the minimum Maven version and added tests in #196 -- looks like we implicitly required 3.3.9 already due to using the latest Maven Enforcer plugin crashing on older versions

mosabua:
Replaced by https://github.com/airlift/airbase/pull/197

PR #194, opened 2019-11-06
Fix lambda opening brace check

findepi:
We want to allow lambdas like

    () -> { throw new RuntimeException(); }


findepi:
cc @martint @mosabua

mosabua:
+1  .. this avoids the failures in the presto build. Hence it also unblocks https://github.com/airlift/airbase/pull/193

PR #196, opened 2019-11-10
Run Travis tests on multiple Maven versions

electrum:


.travis.yml
```diff
@@ -1,12 +1,17 @@
...
-  allow_failures:
-    - jdk: openjdk-ea
+install:
+  - IMAGE=maven:$MAVEN-jdk-$JDK
+  - RUN="run -v $PWD:/build -w /build $IMAGE"
+  - docker pull $IMAGE
+  - docker $RUN mvn -V -B -q dependency:go-offline
+
+script:
+  - docker $RUN mvn -V -B clean install
```
findepi: There is no much building in airbase (as compared to downstream projects), so not sure if it gives us a lot of test coverage.

electrum: That's a good point. Maybe we should add a simple integration test project? But at least this seems better than nothing.

PR #201, opened 2020-01-02
Pin maven-shade-plugin version

kokosing:
Pin maven-shade-plugin version

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -879,6 +879,11 @@
                     <version>3.2</version>
                 </plugin>
 
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-shade-plugin</artifactId>
+                    <version>3.2.1</version>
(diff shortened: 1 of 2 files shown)


airbase/pom.xml
```diff
@@ -879,6 +879,12 @@
                     <version>3.2</version>
                 </plugin>
 
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-shade-plugin</artifactId>
+                    <version>3.1.1</version>
```
electrum: The latest is `3.2.1`, so we should use that.

airbase/pom.xml
```diff
@@ -879,6 +879,12 @@
...
+                </plugin>
+
```
electrum: Nit: no need for a blank line between closing tags

kokosing:
Updated

electrum:
Merged, thanks!

PR #203, opened 2020-01-03
Make surefire heap dump path configurable

sopel39:


airbase/pom.xml
```diff
@@ -110,6 +110,9 @@
         <!-- define the JVM size for forked tests. Defaults to build JVM size. -->
         <air.test.jvmsize>${air.build.jvmsize}</air.test.jvmsize>
 
+        <!-- define the heap dump path for forked tests. -->
+        <air.test.heap-dump-path>/tmp/test-heapdump.hprof</air.test.heap-dump-path>
```
electrum: I think the default is to use the current directory with a unique name. I don’t think we write to a fixed file path and name by default.

sopel39: Yes. I think https://github.com/prestosql/presto/pull/2400 needs to land first. Then we could decide how we upload heap dumps

PR #216, opened 2020-04-21
Add hook for altering surefire JVM argLine

findepi:


findepi:
BTW what does this stand for?
https://github.com/airlift/airbase/blob/087343ca0153fd0f80e2e620c58c6e660f917449/airbase/pom.xml#L434

electrum:
Update CHANGES

electrum:
`${argLine}` is used for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html

findepi:
> `${argLine}` is used for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html

thanks! does it matter we use `$` syntax vs `@{argLine}` in their docs?

PR #218, opened 2020-04-27
Update Guava to 29.0

electrum:


airbase/pom.xml
```diff
@@ -1032,6 +1034,12 @@
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
                 <version>${dep.guava.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>com.google.guava</groupId>
+                        <artifactId>listenablefuture</artifactId>
```
findepi, line 1040: listenablefuture is a guava's dependency and clashes with guava itself?

electrum: I'll add a comment here.

Guava decided to extract `ListenableFuture` to a separate library so that it could be used by Android applications without needed all of Guava. Related, it also pulled out the common `failureaccess` library for "Android reasons". But instead of doing the normal and sane thing of simply extracting the interface, it left the code in Guava, and added a dependency on the dummy `9999.0-empty-to-avoid-conflict-with-guava` version that is simply an empty JAR.

So the empty version depended on here does not conflict, but if a non-empty version somehow got included, it would conflict. We don't technically need to exclude it here, but it's ugly and serves no real purpose (other than working around their stupid choices).

PR #223, opened 2020-05-12
Update to Checkstyle 8.32 and plugin 3.1.1

electrum:


electrum:
There's a (not relevant) security vulnerability in the previous version. Otherwise, looks mostly like bug fixes and new checkers that we could enable.

PR #230, opened 2020-07-03
Exclude guava's annotations dependencies

findepi:
Reference https://github.com/prestosql/presto/blob/03aa1b6f9d9ced2807a33d578384c6e05a6f939a/presto-jdbc/pom.xml#L62-L83:

> ```
> <dependency>
>     <groupId>com.google.guava</groupId>
>     <artifactId>guava</artifactId>
>     <exclusions>
>         <exclusion>
>             <groupId>org.checkerframework</groupId>
>             <artifactId>checker-qual</artifactId>
>         </exclusion>
>         <exclusion>
>             <groupId>com.google.errorprone</groupId>
>             <artifactId>error_prone_annotations</artifactId>
>         </exclusion>
>         <exclusion>
>             <groupId>com.google.j2objc</groupId>
>             <artifactId>j2objc-annotations</artifactId>
>         </exclusion>
>         <exclusion>
>             <groupId>org.codehaus.mojo</groupId>
>             <artifactId>animal-sniffer-annotations</artifactId>
>         </exclusion>
>     </exclusions>
> </dependency>
> ```

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -1037,6 +1037,26 @@
                         <groupId>com.google.guava</groupId>
                         <artifactId>listenablefuture</artifactId>
                     </exclusion>
+                    <!-- exclude optional annotations -->
+                    <exclusion>
+                        <groupId>org.checkerframework</groupId>
+                        <artifactId>checker-qual</artifactId>
+                    </exclusion>
+                    <!-- exclude optional annotations -->
+                    <exclusion>
+                        <groupId>com.google.errorprone</groupId>
+                        <artifactId>error_prone_annotations</artifactId>
+                    </exclusion>
+                    <!-- exclude optional annotations -->
+                    <exclusion>
+                        <groupId>com.google.j2objc</groupId>
+                        <artifactId>j2objc-annotations</artifactId>
+                    </exclusion>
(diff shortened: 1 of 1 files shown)


electrum:
More discussion here: https://github.com/google/guava/issues/2824

findepi:
> The real goal, as I understand, is to prevent these from being packaged for use at runtime. It seems better to attack that problem directly.

Sure.  I do not think we should copy https://github.com/prestosql/presto/blob/03aa1b6f9d9ced2807a33d578384c6e05a6f939a/presto-jdbc/pom.xml#L62-L83 into each of the connectors, but I do think I know the proper way. How should this look like?

electrum:
This seems to be a low priority concern. While it's true that the JARs are not needed at runtime, excluding the would only save a couple hundred kilobytes in total (since they are physically shared across plugins). The reason we exclude them from JDBC is that it's easier to exclude unnecessary dependencies than it is to shade them (plus size is more of a concern there).

findepi:
I am not concerned about size. I know these are small.
Maybe I just incorrectly assumed this would be an improvement.
-- let's continue discussion in https://github.com/prestosql/presto/issues/4336.

PR #233, opened 2020-08-07
Require static imports for Preconditions and Verify

kokosing:
Require static imports for Preconditions and Verify

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@ Airbase 103
 
 * Dependency updates:
   - Joda-Time 2.10.6 (from 2.10.5)
+* Checkstyle updates:
+  - Require static imports for methods from Preconditions and Verify
 
 Airbase 102
 
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -78,6 +78,14 @@
         <property name="format" value="^[ \t]*import com\.google\.common\.base\.MoreObjects;$" />
         <property name="message" value="com.google.common.base.MoreObjects should only be used with static imports" />
     </module>
+    <module name="RegexpMultiline">
+        <property name="format" value="^[ \t]*import com\.google\.common\.base\.Preconditions;$" />
+        <property name="message" value="com.google.common.base.Preconditions should only be used with static imports" />
+    </module>
+    <module name="RegexpMultiline">
+        <property name="format" value="^[ \t]*import com\.google\.common\.base\.Verify;$" />
+        <property name="message" value="com.google.common.base.Verify should only be used with static imports" />
+    </module>
 
     <module name="RegexpMultiline">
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+Airbase 103
+
+* Checkstyle update:
+  - Require static imports for methods from com.google.common.base.Preconditions 
+  and com.google.common.base.Verify
```
findepi, line 5: indent

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -78,6 +78,14 @@
         <property name="format" value="^[ \t]*import com\.google\.common\.base\.MoreObjects;$" />
         <property name="message" value="com.google.common.base.MoreObjects should only be used with static imports" />
     </module>
+    <module name="RegexpMultiline">
+        <property name="format" value="^[ \t]*import com\.google\.common\.base\.Preconditions;$" />
+        <property name="message" value="com.google.common.base.Preconditions should only be used with static imports" />
```
findepi, line 83: The caveat is that a caller class cannot define its own `checkArgument` method and then use Preconditions.
I think I am OK with this.

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+Airbase 103
+
+* Checkstyle update:
```
findepi, line 3:
```suggestion
* Checkstyle updates:
```

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
kokosing: Yes. I would find such custom method very misleading.

kokosing:
Comments addressed. Thanks!

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
martint: This is actually a problem I've run into with String.format. I wanted to have a custom format() method that delegated to String.format, but that kind of usage is blocked. Very annoying.

kokosing: > I wanted to have a custom format() method that delegated to String.format, but that kind of usage is blocked. Very annoying.

I think one might say something opposite when noticed that the `format` got overriden. This surely depends on the case and person.

As with everything this is a tradeoff. https://github.com/prestosql/presto/pull/4733 showed that it was not a big deal in the code base. However, having this checkstyle rule makes a code review process a bit smoother which I find nice. Some people might find nit-picking comments a bit offensive. Thanks to this change we could avoiding discussions on obvious things.

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+Airbase 103
+
+* Checkstyle updates:
+  - Require static imports for methods from com.google.common.base.Preconditions
```
electrum, line 4: Let's shorten this to 

> Require static imports for methods from Preconditions and Verify

kokosing:
Updated.

PR #238, opened 2020-10-06
Require static imports for immutable collection collectors

kokosing:
Require static imports for immutable collection collectors

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -70,6 +70,18 @@
         <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*Math\.toIntExact" />
         <property name="message" value="Math.toIntExact should only be used with static imports" />
     </module>
+    <module name="RegexpSingleline">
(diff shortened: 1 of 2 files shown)


airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -70,6 +70,18 @@
         <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*Math\.toIntExact" />
         <property name="message" value="Math.toIntExact should only be used with static imports" />
     </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*ImmutableMap\.toImmutableMap" />
```
findepi, line 74: I know this is copied from above, but wonder if 
```
^(?!import).* ImmutableMap\.toImmutableMap
```
would be enough.

findepi (approved):

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
kokosing: I don't know. I thought that this way is for a reason.

electrum: I believe that will work, but we can clean that up separately

electrum: Filed https://github.com/airlift/airbase/issues/241 for that cleanup

PR #244, opened 2020-11-01
Upgrade git-commit-id plugin to 4.0.3

wendigo:
This includes fix for https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/524


wendigo:
@electrum We don't need to downgrade this plugin as the new version including jgit fixes was released :)

findepi:
Merged, thanks!

PR #246, opened 2020-11-05
update Guava to 30.0

RaymondWaterlooLi:


kokosing:
In order to verify if it is not a breaking change we could update guava in presto 30.0 and see if it is passing automation. Same thing we need to do for airlift.

wendigo:
Can be closed @kokosing. Guava was updated  in https://github.com/airlift/airbase/commit/47654df9578146d08a43501c98f76b0318b14d56

kokosing:
@findepi Can you please close it?

findepi:
Covered in https://github.com/airlift/airbase/pull/250

@RaymondWaterlooLi sorry that I did not merge yours.
I did not remember about it.

PR #247, opened 2020-11-06
Require static import for String.format

kokosing:
Require static import for String.format

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -71,6 +71,10 @@
         <property name="message" value="Math.toIntExact should only be used with static imports" />
     </module>
     <module name="RegexpSingleline">
+        <property name="format" value="^(?!import ).*String\.format" />
+        <property name="message" value="String.format should only be used with static imports" />
+    </module>
+    <module name="RegexpSingleline">
         <property name="format" value="^(?!import ).*ImmutableMap\.toImmutableMap" />
         <property name="message" value="ImmutableMap.toImmutableMap should only be used with static imports" />
     </module>
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -63,23 +63,23 @@
     </module>
 
     <module name="RegexpSingleline">
-        <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*Objects\.requireNonNull" />
+        <property name="format" value="^(?!import ).*Objects\.requireNonNull" />
         <property name="message" value="Objects.requireNonNull should only be used with static imports" />
     </module>
     <module name="RegexpSingleline">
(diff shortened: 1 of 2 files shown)


skrzypo987 (commented):
That pretty much answers my questions from https://github.com/prestosql/presto/pull/5852

Not giving LGTM here as I am not familiar with those xml files. But looks ok

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -63,23 +63,27 @@
...
         <property name="message" value="Objects.requireNonNull should only be used with static imports" />
     </module>
     <module name="RegexpSingleline">
-        <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*Math\.toIntExact" />
+        <property name="format" value="^(?!import ).*Math\.toIntExact" />
         <property name="message" value="Math.toIntExact should only be used with static imports" />
     </module>
     <module name="RegexpSingleline">
-        <property name="format" value="^([^i]|i[^m]|im[^p]|imp[^o]|impo[^r]|impor[^t]|import[^ ]).*ImmutableMap\.toImmutableMap" />
+        <property name="format" value="^(?!import ).*String\.format" />
```
wendigo, line 74: can we use whole package name here?

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -70,6 +70,10 @@
         <property name="format" value="^(?!import ).*Math\.toIntExact" />
         <property name="message" value="Math.toIntExact should only be used with static imports" />
     </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="^(?!import ).*String\.format" />
```
zaz968m, line 74: This change could be too invasive, especially because another rule exists:
https://github.com/airlift/airbase/blob/fb4d29a27361c6a0a7f11bce8c0b399960bdd62d/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml#L57

Could you please make it a little bit more strict, so that `MyBetter123_String` or `my.Package.String` could have `format` function as well.
So you have to filter out only
`java.lang.String.format` and
`String.format` with `[^a-zA-Z_$0-9])String\.format`

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
zaz968m: I think no, because usually we have package name static imported.

findepi:
> That pretty much answers my questions from [prestosql/presto#5852](https://github.com/prestosql/presto/pull/5852)


Would you mind updating airlift code too?
There are legitimate usecases where String.format cannot be statically imported.

(please be aware there are String.format related changes in https://github.com/airlift/airlift/pull/884)

martint (changes requested):
Airbase is used by many projects. Banning the use of qualified String.format is too draconian. There’s nothing inherently wrong with it (it’s not a source of bugs, etc), other than a code style choice, which is not airbase’s place to be enforcing.

martint (changes requested):
Airbase is used by many projects. Banning the use of qualified String.format is too draconian. There’s nothing inherently wrong with it (it’s not a source of bugs, etc), other than a code style choice, which is not airbase’s place to be enforcing.

PR #255, opened 2021-02-13
Use posix tar mode in assembly-plugin

wendigo:
The difference between `gnu` and `posix` tar formats is explained here: https://www.gnu.org/software/tar/manual/html_section/tar_67.html#Formats

I think that we are good with switch to `posix` tar mode.

cc @electrum @findepi

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -410,8 +410,8 @@
                     <configuration>
                         <!-- must be true for jar-with-dependencies builds -->
                         <appendAssemblyId>true</appendAssemblyId>
-                        <!-- Always use GNU tar mode. -->
-                        <tarLongFileMode>gnu</tarLongFileMode>
+                        <!-- Always use Posix tar mode. -->
+                        <tarLongFileMode>posix</tarLongFileMode>
                     </configuration>
                 </plugin>
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 Airbase 107
 
+* Configure maven-assembly-plugin to use posix tar mode
+
 Airbase 106
 
 * Remove `errorprone-compiler` profile.


findepi (approved):

airbase/pom.xml
```diff
@@ -410,8 +410,8 @@
                     <configuration>
                         <!-- must be true for jar-with-dependencies builds -->
                         <appendAssemblyId>true</appendAssemblyId>
-                        <!-- Always use GNU tar mode. -->
-                        <tarLongFileMode>gnu</tarLongFileMode>
+                        <!-- Always use Posix warn tar mode. -->
+                        <tarLongFileMode>posix_warn</tarLongFileMode>
```
findepi, line 414: what's the difference between `posix` (https://github.com/trinodb/trino/pull/6884) and `posix_warn`?

martint:
What’s the motivation for changing this?

findepi:
> What’s the motivation for changing this?

see https://trinodb.slack.com/archives/CGB0QHWSW/p1613158866364200

martint:
Can we add those details to the commit message for future reference?

airbase/pom.xml
wendigo: this will generate a warning for the paths that exceed maximum length. (default for this plugin is tar mode `warn`): https://github.com/sonatype/plexus-archiver/blob/master/src/main/java/org/codehaus/plexus/archiver/tar/TarArchiver.java

findepi: http://maven.apache.org/plugins/maven-assembly-plugin/single-mojo.html documents this config to matter for lengths over 100 characters.

In Trino, we create an archive with a bunch of files with file name longer than that (note that top dir is part of the tar archive):

```
trino/core/trino-server/target master$ find trino-server-353-SNAPSHOT | sed 's@^./@@' | awk '{print length($0), $0}' | sort -n | tail -11
99 trino-server-353-SNAPSHOT/plugin/hive-hadoop2/google-api-services-storage-v1-rev20190624-1.30.1.jar
100 trino-server-353-SNAPSHOT/plugin/password-authenticators/jetty-alpn-java-client-9.4.30.v20200611.jar
100 trino-server-353-SNAPSHOT/plugin/session-property-managers/jackson-module-parameter-names-2.11.4.jar
102 trino-server-353-SNAPSHOT/plugin/teradata-functions/trino-teradata-functions-353-SNAPSHOT-services.jar
103 trino-server-353-SNAPSHOT/plugin/password-authenticators/trino-password-authenticators-353-SNAPSHOT.jar
103 trino-server-353-SNAPSHOT/plugin/resource-group-managers/trino-resource-group-managers-353-SNAPSHOT.jar
105 trino-server-353-SNAPSHOT/plugin/password-authenticators/http2-http-client-transport-9.4.30.v20200611.jar
107 trino-server-353-SNAPSHOT/plugin/session-property-managers/trino-session-property-managers-353-SNAPSHOT.jar
112 trino-server-353-SNAPSHOT/plugin/password-authenticators/trino-password-authenticators-353-SNAPSHOT-services.jar
112 trino-server-353-SNAPSHOT/plugin/resource-group-managers/trino-resource-group-managers-353-SNAPSHOT-services.jar
116 trino-server-353-SNAPSHOT/plugin/session-property-managers/trino-session-property-managers-353-SNAPSHOT-services.jar
```

* **if** we consider this *OK* from user perspective (for Trino users, and for airlift users in general, they likely have overlapping requirements) -- we should not "warn" here, it's kind of not useful
* **if** we consider this **NOT OK**, we should work towards reducing the file paths lengths

I found similar problems here

- https://hibernate.atlassian.net/browse/HV-1732
- https://github.com/openhab/openhab-distro/pull/434

but I also found someone mentioning drawbacks of the change https://issues.onehippo.com/browse/ARCHE-565?focusedCommentId=498237&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-498237 

> I tried creating tar.gz archives with both Posix and Gnu Problem with the posix tar containing long file names is that the file names could not be extracted on Windows using 7-zip

Overall I am leaning towards thinking we should set this to `posix` (without `_warn`).

wendigo: @findepi I've updated this PR to use posix mode. I'll check if latest windows 7-zip has still this problem with posix tars. Does it matter anyway? Is i.e. Trino on Windows a thing?

wendigo:
Thx @findepi

PR #257, opened 2021-03-28
Allow compiling for Java 16

electrum:


findepi:
_for_ Java 16 or _with_ Java 16?

electrum:
It is “with” since this fixes the tools that read Java 16 bytecode. After this change, the target JDK can be 16.

dain (approved):

PR #259, opened 2021-04-16
Add support for running TestNG and JUnit tests in same module

martint:


airbase/pom.xml
```diff
@@ -437,7 +437,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>2.22.0</version>
+                    <version>3.0.0-M5</version>
```
electrum, line 440: Add a `dep.plugin.surefire.version` property so that these stay the same

electrum (approved):
Update CHANGES

martint:
@electrum , updated

electrum (approved):

PR #262, opened 2021-05-18
Fail build if no tests are executed

Praveen2112:
Allows us to identify false positive cases due to misconfiguration
for tests


electrum:
Good idea. We can explicitly disable for any modules that don't have tests.

PR #266, opened 2021-06-18
Add junit dependency to surefire plugin

martint:
Otherwise, it causes a build failure unless the project includes
it directly (i.e., as a consequence of containing junit-based tests)

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -201,6 +201,7 @@
         <dep.assertj-core.version>3.18.1</dep.assertj-core.version>
         <dep.slice.version>0.39</dep.slice.version>
         <dep.jmh.version>1.20</dep.jmh.version>
+        <dep.junit.version>5.8.0-M1</dep.junit.version>
 
         <!-- license headers -->
         <air.license.owner>${project.organization.name}</air.license.owner>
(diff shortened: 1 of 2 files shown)


airbase/pom.xml
```diff
@@ -465,6 +465,11 @@
                         </argLine>
                     </configuration>
                     <dependencies>
+                        <dependency>
+                            <groupId>org.junit.jupiter</groupId>
+                            <artifactId>junit-jupiter-engine</artifactId>
+                            <version>5.8.0-M1</version>
```
electrum, line 471: We should make the version a property and probably also include the dependencies so they are consistent
```xml
<dependency>
    <groupId>org.junit</groupId>
    <artifactId>junit-bom</artifactId>
    <type>pom</type>
    <version>${dep.junit.version}</version>
    <scope>import</scope>
</dependency>

electrum (approved):
Update CHANGES

airbase/pom.xml
martint: Should that go in the dependencyManagement section?

electrum: Yes. I don't think that `dependencyManagement` affects plugin dependencies, so you probably still need the version property and the usage here

electrum (approved):

PR #275, opened 2021-10-04
Update Guava & Jackson

wendigo:


kokosing (commented):
Can you please try to run Trino Ci for this change? Jackson updates are known for causing issues.

wendigo:
> Can you please try to run Trino Ci for this change? Jackson updates are known for causing issues.

I'll do it but airbase itself is unusable right now due to modernizr update which brings new rules that we don't follow yet

wendigo:
@kokosing it works

kokosing:
Lgtm from me. I cannot merge it as I am not maintainer here. Thank you for checking!

wendigo:
@martint can you review it please?

PR #276, opened 2021-10-04
Update dependencies

wendigo:


wendigo:
@martint wanna take a look?

PR #278, opened 2021-10-06
Update git-commit-id plugin to 4.0.5

wendigo:


wendigo:
This is the last version on the old coordinates. New version (5.0.0) has new coordinates and dropped support for JDK <11. Do we want to drop JDK 8 support as well here @electrum @martint ?

findepi:
Merged, thanks!

PR #283, opened 2021-11-10
Checkstyle version 8.1

Randgalt:
Checkstyle v 9.1 fixes incorrect errors on switch expressions that
assign to a variable. E.g.

```
String s = switch(foo) {
    case A -> "it's a";
    default -> "it's not a";
};
```

The current version generates an error for this.


martint:
Merged as 69c1b58692afd618f9f263555338b2340f881f12

PR #289, opened 2021-12-29
Allow modules without tests or that don't use TestNG

electrum:


airbase/pom.xml
```diff
@@ -1391,6 +1389,33 @@
...
+                    <value>true</value>
+                </property>
+            </activation>
+            <build>
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-surefire-plugin</artifactId>
+                            <dependencies>
```
martint, line 1406: Does this override the configuration above to use JUnit? If so, this will break projects that use both frameworks.

electrum: Any project that uses both will need to include both explicitly. Surefire will detect JUnit or TestNg, but not both, unfortunately. The problem with including both in the base POM is that any project without TestNG on the classpath will fail. Ideally, the TestNG provider would be skipped if TestNG is not present, but instead it fails with `ClassNotFoundException`. There's no good way to conditionally add the dependency, since you can't active a profile with a Maven property, and there is no way to exclude it in a child project, since the dependencies section doesn't use `combine.self="override"` like configuration does.

electrum: The downside is that if you have both, but don't include the providers, the TestNG tests will be silently skipped. But that seems better than forcing every module that doesn't use TestNG to include TestNG on the test classpath.

electrum: I fixed the commits and added a description

airbase/pom.xml
```diff
@@ -452,7 +454,7 @@
                         </systemPropertyVariables>
                         <trimStackTrace>false</trimStackTrace>
                         <runOrder>random</runOrder>
-                        <failIfNoTests>true</failIfNoTests>
+                        <failIfNoTests>${air.test.fail-if-no-tests}</failIfNoTests>
```
findepi, line 457: I am wondering what's the usecase for this would be.

If module has no tests, there is no point in invoking surefire at all.
So I'd want to define `air.skip-tests` or `air.there-are-no-tests` property instead.

If, however, the module has tests, but i want to skip test on per-execution basis, the `-DskipTests` is there.

Thus, we shouldn't add this property as is.

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 Airbase 117
 
 * Add `air.test.fail-if-no-tests` property
+* Remove test providers from Surefire plugin
```
findepi, line 4: "Fix build failure when there are Junit tests and no TestNG tests"
also add a note about a backward incompatible change

> However, if a
module uses both, only the JUnit tests will run -- the TestNG tests
will be silently skipped.

this is gonna be a problem for Trino, as there is a junit test there already.
With this change, we will silently looe test coverage.
Is there any way to either keep test providers but avoid failure, or have a check for mixed junit+testng modules?
Maybe we could have our own version of these providers?

The failure is:
```
java.lang.NoClassDefFoundError: org/testng/TestNG
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:87)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:112)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:99)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:145)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:428)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:162)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:562)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:548)
Caused by: java.lang.ClassNotFoundException: org.testng.TestNG
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:606)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:168)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 8 more
```

so maybe it would be sufficient to have an activator that triggers on "has testng on classpath" condition.

findepi (changes requested):

CHANGES.md
martint: Isn’t that as simple as configuring the surefire plugin with the necessary providers directly in Trino when we update to this version of airbase?

airbase/pom.xml
electrum: Good point, we should just skip the plugin entirely for such modules.

CHANGES.md
electrum: Yes, any modules with mixed tests need to declare both. If we consider mixed to be “transitional”, with a goal of migrating fully from one to the other, then this seems reasonable.

Maybe @jvanzyl could get upstream Surefire fixed to work automatically with modules having both TestNG and JUnit 5?

dain (approved):

CHANGES.md
findepi: > Isn’t that as simple as configuring the surefire plugin with the necessary providers directly in Trino

That might do, if we assume Trino is the only downstream consumer of the project.

martint: It’s not, but if this is documented as a behavior change, it should be a hint for users to update their project accordingly.

airbase/pom.xml
```diff
@@ -450,6 +452,7 @@
                             <java.awt.headless>true</java.awt.headless>
                             <java.util.logging.SimpleFormatter.format>%1$tY-%1$tm-%1$tdT%1$tH:%1$tM:%1$tS.%1$tL%1$tz %4$s %5$s%6$s%n</java.util.logging.SimpleFormatter.format>
                         </systemPropertyVariables>
+                        <skipTests>${air.test.skip}</skipTests>
```
findepi, line 455: When `air.test.skip` is true, we could perhaps exclude the whole `maven-surefire-plugin`, could we?
re-sharing https://peter.palaga.org/presentations/211021-baselone-maven-my-life-is-short/ shared with me by  @nineinchnick

PR #294, opened 2022-01-14
Checkstyle fix for missing violation about missing space after switch(...)

ksobolew:
We override the list of tokens in `WhitespaceAround` in order to add some more, but `LCURLY` was missing relative to the upstream's list. This fixes a missing checkstyle violation after a switch, probably among other things:

```java
switch (something){
    case CASE: return result();
}
```

Not sure why all statements except `switch` were checked correctly.

The `allowEmptyTypes` option needs to be added so that empty anonymous types do not cause violation because of missing space before `LCURLY`:

```java
doSomethingWith(Key.get(new TypeLiteral<MyType>() {}, AnnotatedWith.class)); // this is fine
```


ksobolew:
cc: @findepi

kokosing:
Have you run it against Trino repo to see how it works there?

ksobolew:
> Have you run it against Trino repo to see how it works there?

Yes, there's only a bunch of trivial things like the following in a single class:

```diff
index 78b8187a65..58e837a26a 100644
--- a/plugin/trino-http-event-listener/src/test/java/io/trino/plugin/httpquery/TestHttpQueryListener.java
+++ b/plugin/trino-http-event-listener/src/test/java/io/trino/plugin/httpquery/TestHttpQueryListener.java
@@ -258,7 +258,7 @@ public class TestHttpQueryListener
     public void testContentTypeDefaultHeaderShouldAlwaysBeSet()
             throws Exception
     {
-        EventListener querylogListener = factory.create(new HashMap<>(){{
+        EventListener querylogListener = factory.create(new HashMap<>() {{
                 put("http-event-listener.connect-ingest-uri", server.url("/").toString());
                 put("http-event-listener.log-completed", "true");
             }});
```

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -213,11 +213,12 @@
             <property name="allowEmptyConstructors" value="true" />
             <property name="allowEmptyMethods" value="true" />
             <property name="allowEmptyLambdas" value="true" />
+            <property name="allowEmptyTypes" value="true" />
```
findepi, line 216: > The `allowEmptyTypes` option needs to be added so that empty anonymous
> types do not cause violation because of missing space before `LCURLY`:
> ```
> doSomethingWith(Key.get(new TypeLiteral<MyType>() {}, AnnotatedWith.class)); // this is fine
> ```

Your example _does_ have a space before left left curly. What's the issue?

findepi:
> `new HashMap<>(){{`

the `{{` is an anti-pattern. use `Set.of` instead.

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
ksobolew: It's "whitespace *around*", so it would also complain about missing space after the left curly.

ksobolew:
> the `{{` is an anti-pattern. use `Set.of` instead.

Good thing that checkstyle found it, then :) I think there targeted checks for this pattern, too.

findepi:
> > Have you run it against Trino repo to see how it works there?
> 
> Yes, there's only a bunch of trivial things like the following in a single class:


Any outstanding issues after https://github.com/trinodb/trino/pull/10633 merged?

ksobolew:
> Any outstanding issues after [trinodb/trino#10633](https://github.com/trinodb/trino/pull/10633) merged?

No issues, checkstyle is clean.

findepi:
@ksobolew please add CHANGES entry

PR #299, opened 2022-01-25
Update guice to 5.1.0

wendigo:
This version brings compatibility with JDK17+

See: https://github.com/google/guice/wiki/Guice510


wendigo:
cc @martint @kokosing

PR #300, opened 2022-02-25
Define default `argLine` property

nineinchnick:
Define an empty default value for the `argLine` property, required when `air.check.skip-jacoco` is true. This was introduced in the 120 version, where unused plugins are disabled completely.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 121
+
+* Fix Surefire setup when JaCoCo is disabled
+
 Airbase 120
 
 * Add dependencies:
@@ -9,7 +13,7 @@ Airbase 120
   - Joda-Time 2.10.13 (from 2.10.10)
 * Plugin updates:
   - update maven-compiler-plugin 3.9.0 (from 3.8.1)
-  
+
 Airbase 119
 
 * Remove `air.test.skip` property. Use `skipTests` instead.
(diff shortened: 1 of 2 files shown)


nineinchnick:
CC @findepi

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+Airbase 121
+
+* Define a default for argLine property, required when `air.check.skip-jacoco` is true
```
findepi, line 3: Maybe "Fix Surefire setup when JaCoCo is disabled"?

airbase/pom.xml
```diff
@@ -225,6 +225,9 @@
         <air.repository.release.id>ossrh</air.repository.release.id>
         <air.repository.release.name>Sonatype Release Snapshots</air.repository.release.name>
         <air.repository.release.url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</air.repository.release.url>
+
+        <!-- ${argLine} is defined by Jacoco, but it can be disabled, so define an empty default here -->
+        <argLine />
```
findepi, line 230: Can this be in a profile?

nineinchnick: This is required by surefire which is always enabled. Do you see other benefits for moving it to a profile?

findepi: Wi thought we can define `argLine` when JaCoCo is not loaded (in a profile), and let JaCoCo define argLine when it is loaded (in a profile).

nineinchnick: Great idea, we shouldn't be worried about compatibility with Jacoco this way. Done and it works!

PR #303, opened 2022-03-18
Allow configuring if existing headers should be skipped in license check

hashhar:
All credit to @findinpath for finding this.

See for source https://github.com/trinodb/trino/pull/11563#issuecomment-1072302541

cc: @findinpath @findepi

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -210,6 +210,7 @@
         <air.license.default-value>Copyright (C) ${project.inceptionYear} ${air.license.owner}</air.license.default-value>
         <air.license.ensure-match>Copyright \(C\) \d{4} .+</air.license.ensure-match>
         <air.license.header-file>license/apache-header.txt</air.license.header-file>
+        <air.license.skip-existing-headers>false</air.license.skip-existing-headers>
 
         <!-- Checkstyle -->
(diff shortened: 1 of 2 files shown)


airbase/pom.xml
```diff
@@ -738,7 +739,7 @@
                     </dependencies>
                     <configuration>
                         <skip>${air.check.skip-license}</skip>
-                        <skipExistingHeaders>true</skipExistingHeaders>
+                        <skipExistingHeaders>${air.license.skip-existing-headers}</skipExistingHeaders>
```
findepi, line 742: Why not always `false`?

findepi: (it's `true` since ever (d71d25e711593aff2a1e339d146d8a83205e2996))

hashhar: Just to preserve compatibility. IMO it should be false.

findepi: @electrum can you please chime in here?

electrum: Let’s make this false, since that is the better default. Any projects that need to skip can set the property when updating.

hashhar: Updated default to true and adjusted CHANGES entry accordingly.

findepi (approved):

PR #310, opened 2022-04-07
Update Surefire plugin

nineinchnick:
The new version includes https://github.com/apache/maven-surefire/pull/407 which should reduce memory usage in CI and help avoid failures caused by running out of memory.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@ Airbase 124
 
 * Dependency updates:
   - Slice 0.42 (from 0.41)
+* Plugin updates:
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,8 @@
+Airbase 125
```
findepi, line 1: we're still at 124

findepi:
thanks, i didn't notice the release

CHANGES.md
nineinchnick: Done!

findepi (approved):

PR #311, opened 2022-05-13
Disallow shaded org.apache imports

hashhar:
Lot of dependencies shade Apache code and it's easy to import from them
by mistake which often breaks once dependencies are updated and have
improved their exclusions/shading.

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -247,6 +247,8 @@
             <!-- Shaded okhttp3; match "shaded", since "com.example.okhttp3" could be OkHttp-specific code within com.example -->
             <property name="illegalPkgs" value=".*\.shaded\.okhttp3" />
             <property name="regexp" value="true" />
+            <!-- Shaded org.apache code -->
+            <property name="illegalPkgs" value=".*\.shaded\.org\.apache" />
         </module>
 
         <module name="MethodName">
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 Airbase 126
 
 * Fix the way locale information is passed to maven-surefire-plugin
+* Checkstyle updates:
+  - Forbid imports of shaded Apache code
 * Dependency updates:
   - Logback 1.2.11 (from 1.2.3)
 * Plugin downgrades:


hashhar:
cc: @findepi @wendigo

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -251,6 +251,8 @@
             <!-- Shaded okhttp3; match "shaded", since "com.example.okhttp3" could be OkHttp-specific code within com.example -->
             <property name="illegalPkgs" value=".*\.shaded\.okhttp3" />
             <property name="regexp" value="true" />
+            <!-- Shaded commons-lang -->
+            <property name="illegalPkgs" value=".*\.org\.apache\.commons\.lang" />
```
kokosing, line 255:
```
            <property name="illegalPkgs" value=".*\.org\.apache" />
```
?

hashhar: It may cause issues with some Apache projects like Kafka and HBase/Phoenix which use shaded versions of some org.apache code. 

If left to me I'd prefer to disallow any shaded import and instead and exclusions where needed.

kokosing: This is what I have found in Trino:
```
kokosing@m16:~/trino$ git grep 'import ' | grep '\.org\.apache'
core/trino-main/src/test/java/io/trino/operator/aggregation/listagg/TestListaggAggregationFunction.java:import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
core/trino-main/src/test/java/io/trino/sql/query/TestListagg.java:import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
plugin/trino-phoenix/src/test/java/io/trino/plugin/phoenix/TestingPhoenixServer.java:import org.apache.phoenix.shaded.org.apache.zookeeper.server.ZooKeeperServer;
plugin/trino-phoenix5/src/test/java/io/trino/plugin/phoenix5/TestingPhoenixServer.java:import org.apache.phoenix.shaded.org.apache.zookeeper.server.ZooKeeperServer;
```

The `ZooKeeperServer` is imported only for changing the logging so we could skip import there.

findepi: @hashhar would it work if we disallow `.*\.shaded\.org\.apache`?  (adding "shaded", but removing "cmmons.lang")
see okhttp3 case above.

> some Apache projects like Kafka and HBase/Phoenix which use shaded versions of some org.apache code.

in their API?

findepi:
cc @electrum

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
hashhar: I checked and it seems it's possible to do this (would require some changes in Trino).

Marking this as draft until I have a fix in Trino ready.

wendigo (approved):

electrum (approved):

hashhar:
@electrum Rebased to resolve conflicts (CHANGES.md). Please merge.

PR #315, opened 2022-05-25
Fix the way locale information is passed to Surefire

ksobolew:
This moves primarily the `user.language` and `user.region` from `systemPropertyVariables` to explicit property setting on the JVM. These two methods to set them are supposed to be the same, but passing via system properties does not actually work. Which is annoying, because when the are set in both ways, Surefire complains about it:

```
[WARN] The system property user.language is configured twice! The property appears in <argLine/> and any of <systemPropertyVariables/>, <systemProperties/> or user property.
[WARN] The system property user.region is configured twice! The property appears in <argLine/> and any of <systemPropertyVariables/>, <systemProperties/> or user property.
```

It seems that the `user.timezone` actually works as expected, but is moved along with the other two for consistency.

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -426,9 +426,6 @@
                     <configuration>
                         <systemPropertyVariables>
                             <sun.jnu.encoding>${project.build.sourceEncoding}</sun.jnu.encoding>
-                            <user.timezone>${air.test.timezone}</user.timezone>
-                            <user.language>${air.test.language}</user.language>
-                            <user.region>${air.test.region}</user.region>
                             <java.awt.headless>true</java.awt.headless>
                             <java.util.logging.SimpleFormatter.format>%1$tY-%1$tm-%1$tdT%1$tH:%1$tM:%1$tS.%1$tL%1$tz %4$s %5$s%6$s%n</java.util.logging.SimpleFormatter.format>
                         </systemPropertyVariables>
@@ -440,6 +437,13 @@
                         <!-- ${argLine} is for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html -->
                         <argLine>
                             ${argLine}
+
+                            <!-- Pass the following three directly to the JVM,
+                                 because systemPropertyVariables are not passed correctly by Maven -->
+                            -Duser.timezone=${air.test.timezone}
+                            -Duser.language=${air.test.language}
+                            -Duser.region=${air.test.region}
+
                             -Dfile.encoding=${project.build.sourceEncoding}
                             -Xmx${air.test.jvmsize}
                             -Xms${air.test.jvmsize}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 126
+
+* Fix the way locale information is passed to maven-surefire-plugin
+
 Airbase 125
 
 * Checkstyle updates:


airbase/pom.xml
```diff
@@ -447,6 +445,8 @@
                             -XX:+HeapDumpOnOutOfMemoryError
                             -XX:-OmitStackTraceInFastThrow
                             ${air.test.jvm.additional-arguments}
+                            -Duser.language=${air.test.language}
+                            -Duser.region=${air.test.region}
```
findepi, line 449: keep `air.test.jvm.additional-arguments` last

findepi:
> These two are supposed to be the same, but passing via system properties does not actually work.

See https://github.com/airlift/airbase/pull/98#issuecomment-420022274

Does this mean #98 never actually worked?
@findepi reported https://github.com/prestodb/presto/issues/9313#issuecomment-457174421 it did work.

@ksobolew how do you test these changes?

ksobolew:
> @ksobolew how do you test these changes?

I just run `./mvnw verify -pl :trino-main -Dtest=TestRealOperators` in Trino. My locale is pl_PL, so I get test failures of the sort: `java.lang.AssertionError: expected [7.541985E2] but found [7,541985E2]`, which clearly indicates that the system default locale was used instead of the one specified in the Surefire config.

> @findepi reported https://github.com/prestodb/presto/issues/9313#issuecomment-457174421 it did work.

Maybe if broke in the meantime? That configuration was added 4 years ago... I did some googling and found problem reports about that, but no fix in Surefire was ever mentioned.

findepi:
Thanks for testing this out.
Do we also need to move the time zone?
Perhaps not, for the timezone we have https://github.com/findepi/trino/blob/7fd62db50aec490907e93412441fe65dbfea6052/testing/trino-tests/src/test/java/io/trino/tests/TestVerifyTrinoTestsTestSetup.java#L28.

Does this test pass for you locally?

(even if it does, i would somewhat prefer to set timezone region and lang same way)

cc @electrum

ksobolew:
> Thanks for testing this out. Do we also need to move the time zone? Perhaps not, for the timezone we have https://github.com/findepi/trino/blob/7fd62db50aec490907e93412441fe65dbfea6052/testing/trino-tests/src/test/java/io/trino/tests/TestVerifyTrinoTestsTestSetup.java#L28.

Maybe we need similar test for the locale?

> Does this test pass for you locally?

Yes, this makes the tests pass for me.

ksobolew:
> Maybe we need similar test for the locale?

See https://github.com/trinodb/trino/pull/12560/files#diff-30ea3cddc958abc26c37e93c555a0f8cbc422ca55e7b1cc907a1e1aea72fcd6fR33

ksobolew:
Addressed comments. One interesting thing to note is that this is actually broken in Maven, but IntelliJ correctly passes these parameters so these test work fine in the IDE.

airbase/pom.xml
electrum: We should put those above with `-Dfile.encoding=`

ksobolew: Done!

airbase/pom.xml
```diff
@@ -440,6 +437,9 @@
                         <!-- ${argLine} is for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html -->
                         <argLine>
                             ${argLine}
+                            -Duser.timezone=${air.test.timezone}
```
findepi, line 440: add a code comment why this goes in argsLine and not systemPropertyVariables

ksobolew: Added :)

findepi:
As a result of this change, my IntelliJ no longer sets `-Duser.timezone=${air.test.timezone}` when running tests, so they fail to start because of checks like
https://github.com/trinodb/trino/blob/4aacac24cb2811970abaa5acfa47427206568562/plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java#L238
or 
https://github.com/trinodb/trino/blob/b2fb2b5a88e7452348e8f67d379fb555d3c49a59/plugin/trino-postgresql/src/test/java/io/trino/plugin/postgresql/TestPostgreSqlTypeMapping.java#L153

@ksobolew does it work on your side?

ksobolew:
Hm, I haven't encountered any problems 🤔

ksobolew:
Not yet, anyway

findepi:
> Hm, I haven't encountered any problems 🤔

Do Trino's `TestPostgreSqlTypeMapping` and `TestHiveConnectorTest` work for you?

ksobolew:
Looks like IDEA is confused by `${argLine}` in Airbase's surefire configuration:

```xml
                        <argLine>
                            ${argLine}
                            ...
                        </argLine>
```

Once I removed it, it worked. Thing is, if there's no Jacoco, this does not get expanded and sits there as literal `${argLine}`.

nineinchnick:
See https://github.com/airlift/airbase/pull/300 that should provide a default value when Jacoco is disabled

nineinchnick:
Would definining the properties twice, both in `systemPropertyVariables` and `argLine` work?

findepi:
> See #300 that should provide a default value when Jacoco is disabled

Seems like it was merged long ago, and IntelliJ doesn't pick the time zone ~today.

> Would definining the properties twice, both in systemPropertyVariables and argLine work?

would prefer to avoid duplication, but maybe we need to do this

this will make surefire issue warnings, like mentioned in the PR description

> These two methods to set them are supposed to be the same, but passing via system properties does not actually work. Which is annoying, because when the are set in both ways, Surefire complains about it: [...]

> It seems that the user.timezone actually works as expected, but is moved along with the other two for consistency.

or maybe we move back the user.timezone (provided IntelliJ can pick the other to)

BTW @ksobolew is there a surefire bug report about this?

ksobolew:
> See #300 that should provide a default value when Jacoco is disabled

It was then changed in #301 to activate on `air.check.skip-all` instead of `air.check.skip-jacoco` (for a valid reason that it didn't work). But in Trino we explicitly set `<air.check.skip-jacoco>true</air.check.skip-jacoco>` so the profile `jacoco-check` still gets activated. So `jacoco-defaults` does not and `argLine` is not set.

ksobolew:
> BTW @ksobolew is there a surefire bug report about this?

Actually with Surefire it works just fine

ksobolew:
> Would definining the properties twice, both in `systemPropertyVariables` and `argLine` work?

Not sure, probably yes. But there are two "but"s:

* Specifying both is not an error, but the runner logs warnings about it (not fatal, so we can live with it)
* This actually prevents passing *all* additional arguments to IDEA (though probably not all are useful when running in the IDE)

nineinchnick:
Can Trino define a default for `argLine` if it has Jacoco  always disabled?

ksobolew:
> Can Trino define a default for `argLine` if it has Jacoco always disabled?

I think it can. This will make enabling jacoco harder, though.

Can we activate the jacoco profiles explicitly?

ksobolew:
> Was _that_ problematic behavior reported?

Oh, right :) I guess not. The best I could find was https://stackoverflow.com/questions/8901880/set-surefire-plugin-locale-in-maven-pom-file

> The bug is in airbase. In trino we can workaround, sure, but can we fix this here?

Fair question. I hope so!

findepi:
> > Was _that_ problematic behavior reported?
> 
> Oh, right :) I guess not. T

We should have reported this before/while working on this PR. Can you please follow-up accordingly?
(Also, maybe it's easy to fix in the surefire plugin.)

ksobolew:
Reported it as https://issues.apache.org/jira/browse/SUREFIRE-2104

ksobolew:
> > This actually prevents passing all additional arguments to IDEA (though probably not all are useful when running in the IDE)
> 
> Can you expand on this? What would be worse than it is today, or before this PR was merged?

The presence of the unexpanded `${argLine}` causes IDEA to not include anything that's in the `argLine` configuration element. So this is a pre-existing issue, we only started to see it "thanks" to these two properties.

findepi:
> Can Trino define a default for `argLine` if it has Jacoco always disabled?

I was not successful in making this work.



> IntelliJ no longer sets `-Duser.timezone=${air.test.timezone}` when running tests, so they fail to start because of checks like
> https://github.com/trinodb/trino/blob/4aacac24cb2811970abaa5acfa47427206568562/plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java#L238
> or
> https://github.com/trinodb/trino/blob/b2fb2b5a88e7452348e8f67d379fb555d3c49a59/plugin/trino-postgresql/src/test/java/io/trino/plugin/postgresql/TestPostgreSqlTypeMapping.java#L153

is it only for me, or is this problem persistent?
@ksobolew how did you solve this on your side?

ksobolew:
> > Can Trino define a default for `argLine` if it has Jacoco always disabled?
> 
> I was not successful in making this work.
> 
> > IntelliJ no longer sets `-Duser.timezone=${air.test.timezone}` when running tests, so they fail to start because of checks like
> > https://github.com/trinodb/trino/blob/4aacac24cb2811970abaa5acfa47427206568562/plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java#L238
> > or
> > https://github.com/trinodb/trino/blob/b2fb2b5a88e7452348e8f67d379fb555d3c49a59/plugin/trino-postgresql/src/test/java/io/trino/plugin/postgresql/TestPostgreSqlTypeMapping.java#L153
> 
> is it only for me, or is this problem persistent? @ksobolew how did you solve this on your side?

It was some time ago and I don't remember exactly what the problem was, but I checked now and indeed, it doesn't work. In fact, it seems that IDEA does not pick *any* of the arguments specified inside `argLine`. Reverting this commit at least makes the timezone to be passed correctly, which is the opposite of what seemingly was the original issue.

findepi:
> Reverting this commit at least makes the timezone to be passed correctly

awesome, let's do this, thanks!

> which is the opposite of what seemingly was the original issue.

the PR title talks about locale

ksobolew:
> and they have the same issues

I read the original commit message and it seems that it's wrong - they are *not* handled the same way.

ksobolew:
No, I don't think there is anything that prevents us, except for making sure that it works.

ksobolew:
@findepi did the partial revert in airlift/airbase#358. Would be cool if you could verify locally if it fixes the issue for you (I did some tests and it seems it does for me).

PR #316, opened 2022-06-07
Update Surefire plugin

nineinchnick:
This is another attempt at #310 which had to be reverted in #313 because of https://issues.apache.org/jira/browse/SUREFIRE-2068

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+Airbase 128
+
+* Plugin updates:
+  - Surefire 3.0.0-M7 (from 3.0.0-M5)
+
 Airbase 127
 
 * Checkstyle updates:
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -180,7 +180,7 @@
 
         <!-- Plugin versions used in multiple places -->
         <dep.plugin.scm.version>1.8.1</dep.plugin.scm.version>
-        <dep.plugin.surefire.version>3.0.0-M5</dep.plugin.surefire.version>
+        <dep.plugin.surefire.version>3.0.0-M7</dep.plugin.surefire.version>
 
         <!-- Packaging for the tarball deployment -->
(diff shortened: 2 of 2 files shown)


nineinchnick:
> Mentioned bug is still open.

Right, it's not in the release notes either: https://github.com/apache/maven-surefire/releases/tag/surefire-3.0.0-M7

But there's a comment in the issue that it was supposed to be fixed in master in April. I'll test this latest version in Trino anyway.

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+Airbase 127
```
nineinchnick, line 1: 126 hasn't been released yet, so I need to replace the `Plugin downgrades` section in Airbase 126.

findepi: Maybe we should release 126 first, so that we have the other changes delivered to airbase users?

we should test the surefire upgrade on the trino, ideally before merging this pr

nineinchnick: So I ran tests on trino-main more than 10 times today (it takes about 15 minutes) and everything was fine. I did manage to reproduce the issue just once, after running it about 6 times, with airbase 125.

findepi: I published a SNAPSHOT release with this PR as https://oss.sonatype.org/content/repositories/snapshots/io/airlift/airbase/126-SNAPSHOT/airbase-126-20220608.202754-1.pom

you can run full Trino CI with this.

nineinchnick: Great, thanks! https://github.com/trinodb/trino/pull/12758

nineinchnick: I rebased and updated this file.

nineinchnick:
`test-main` failed with:
```
Terminating due to java.lang.OutOfMemoryError: Java heap space
```
https://github.com/trinodb/trino/runs/6809221471?check_suite_focus=true

PR #318, opened 2022-06-22
Always declare empty argLine

ksobolew:
The `argLine` property is used by Jacoco to store the JVM commands which enable its agent. It is initialized by the Jacoco plugin, then used in the Surefire configuration. But when Jacoco is disabled (e.g. using the `air.check.skip-jacoco` property), the `argLine` property is not initialized, and if we don't take additional measures, it will appear as literal `${argLine}` in Surefire's JVM arguments line, which may cause trouble with some tools. The way it was handled previously was that there was an additional profile, which was supposed to be activated when Jacoco was disabled, and which initialized the `argLine` property to empty, allowing Maven to properly interpolate it in the Surefire configuration.

There was one problem with this solution, though - these are system properties, not Maven properties, and profile activation is done before property values are resolved. So if we have a project which disables Jacoco explicitly in its POM (by overriding the property like this: `<air.check.skip-jacoco>true</air.check.skip-jacoco>`), this did not change the set of active profiles. This means that the Jacoco plugin was still configured - though disabled - and the prifile to initialize the `argLine` property was still not active. Thus, when the Surefire configuration was evaluated, the `argLine` property was uninitialized.

The fix in this commit unconditionally initializes `argLine` to empty. This way if Jacoco is enabled, the Jacoco plugin will overwrite this property; otherwise it will remain empty. But for this to work we have to change the Surefire configuration to use late evaluation for `argLine` to prevent its evaluation before it's reinitialized by the Jacoco plugin.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@ Airbase 128
 
 * Plugin updates:
   - JaCoCo 0.8.8 (from 0.8.7)
+* Always define ``argLine`` property even if Jacoco is disabled, to ensure
+  that Surefire configuration does not confuse IntelliJ IDEA
 
 Airbase 127
 
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -178,6 +178,10 @@
         <!-- See http://stackoverflow.com/questions/1012402/maven2-property-that-indicates-the-parent-directory    -->
         <air.main.basedir>${project.basedir}</air.main.basedir>
 
+        <!-- ${argLine} is used by Jacoco to store commands to enable its agent, and then used in Surefire configuration.
+             When Jacoco is not enabled, it will stay empty, otherwise it will be overwritten. -->
+        <argLine />
+
         <!-- Plugin versions used in multiple places -->
         <dep.plugin.scm.version>1.8.1</dep.plugin.scm.version>
         <dep.plugin.surefire.version>3.0.0-M5</dep.plugin.surefire.version>
@@ -434,9 +438,9 @@
                         <failIfNoTests>true</failIfNoTests>
                         <parallel>${air.test.parallel}</parallel>
                         <threadCount>${air.test.thread-count}</threadCount>
-                        <!-- ${argLine} is for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html -->
+                        <!-- @{argLine} is for Jacoco: https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html -->
                         <argLine>
-                            ${argLine}
+                            @{argLine}
 
                             <!-- Pass the following three directly to the JVM,
                                  because systemPropertyVariables are not passed correctly by Maven -->
@@ -1372,20 +1376,6 @@
                     </plugin>
                 </plugins>
             </build>
-        </profile>
-        <profile>
-            <id>jacoco-defaults</id>
-            <activation>
-                <property>
-                    <name>air.check.skip-all</name>
-                    <value>true</value>
-                </property>
-            </activation>
-            <properties>
-                <!-- ${argLine} is defined by Jacoco, but required by Surefire, so define an empty default
-                when Jacoco is disabled with all checks -->
-                <argLine />
-            </properties>
         </profile>
         <profile>
             <id>modernizer-check</id>


findepi:
@ksobolew please update CHANGES.

@nineinchnick PTAL

airbase/pom.xml
```diff
@@ -178,6 +178,10 @@
         <!-- See http://stackoverflow.com/questions/1012402/maven2-property-that-indicates-the-parent-directory    -->
         <air.main.basedir>${project.basedir}</air.main.basedir>
 
+        <!-- ${argLine} is used by Jacoco to store commands to enable its agent, and then used in Surefire configuration.
+             When Jacoco is not enabled, it will stay empty, otherwise it will be overwritten. -->
```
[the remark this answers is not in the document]
ksobolew: Yes, I did. I know it works because in Trino Jacoco complains about too-new bytecode version (I'm running Java 18) :)

```
$ mvnd -pl :trino-tests test -Dtest=TestVerifyTrinoTestsTestSetup -Dair.check.skip-jacoco=false

...

[INFO] --- jacoco-maven-plugin:0.8.7:prepare-agent (default) @ trino-tests ---
[INFO] argLine set to -javaagent:/Users/kudivuhadi/.m2/repository/org/jacoco/org.jacoco.agent/0.8.7/org.jacoco.agent-0.8.7-runtime.jar=destfile=/private/var/tmp/trino/testing/trino-tests/target/jacoco.exec

...

[INFO] --- maven-surefire-plugin:3.0.0-M5:test (default-test) @ trino-tests ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running io.trino.tests.TestVerifyTrinoTestsTestSetup
[WARN] [stderr] 2022-06-22T09:13:14.490-0500 INFO LogTestDurationListener enabled: false
[WARN] [stderr] 2022-06-22T09:13:14.517-0500 INFO ProgressLoggingListener is disabled!
[WARN] [stderr] java.lang.instrument.IllegalClassFormatException: Error while instrumenting sun/util/resources/cldr/provider/CLDRLocaleDataMetaInfo.
...
[WARN] [stderr] Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 62
...
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.427 s - in io.trino.tests.TestVerifyTrinoTestsTestSetup
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] Segment walltime 3 s, segment projects service time 3 s, effective/maximum degree of concurrency 1,00/15
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.893 s (Wall Clock)
[INFO] Finished at: 2022-06-22T16:13:14+02:00
[INFO] ------------------------------------------------------------------------
```

nineinchnick (approved):

CHANGES.md
```diff
@@ -2,6 +2,7 @@ Airbase 128
 
 * Plugin updates:
   - JaCoCo 0.8.8 (from 0.8.7)
+* Fix configuration of Surefire when Jacoco is disabled via property in POM
```
findepi, line 5: I think we're fixing IntelliJ's behavior

ksobolew: Sure, but who knows what other tools are confused by this as well...

nineinchnick:
LGTM

CHANGES.md
findepi: well, we know that Surefire maven plugin is not confused, so we shouldn't say we're fixing "configuration for Surefire".

nineinchnick: How about something like this?
```suggestion
* Ensure ``argLine`` is always defined, with or without Jacoco being enabled
```

ksobolew: I'll work for me :) The details are in the commit message (another Leo Tolstoy badge guaranteed)

CHANGES.md
```diff
@@ -2,6 +2,7 @@ Airbase 128
 
 * Plugin updates:
   - JaCoCo 0.8.8 (from 0.8.7)
+* Ensure ``argLine`` is always defined, with or without Jacoco being enabled
```
findepi, line 5: Or maybe 

```
* Always define ``argLine`` regardless whether Jacoco is enabled to help IntelliJ IDEA
  tests configuration from Surefire maven plugin's configuration.
```

ksobolew: I feel like we're deep into bikeshed territory here ;) I'm kind of reluctant to mention IDEA specifically here, but not exactly sure why. But I don't have strong feelings about it so why not.

ksobolew: How about now?

* Always define ``argLine`` property even if Jacoco is disabled, to ensure
  that Surefire configuration does not confuse IntelliJ IDEA

findepi (approved):

PR #319, opened 2022-07-24
Update dependencies: slf4j, joda-time

nineinchnick:
Update slf4j to be able to update testcontainers to the latest version, since it also depends on slf4j.

Update joda-time to the latest version. This brings back some timezones removed earlier: https://www.joda.org/joda-time/changes-report.html#a2.10.14

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 Airbase 129
 
 * Dependency updates:
+  - joda-time 2.10.14 (from 2.10.13)
   - slf4j 1.7.36 (from 1.7.32)
 
 Airbase 128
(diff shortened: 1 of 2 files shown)


kokosing:
Should we define testcontainers here too?

nineinchnick:
> Should we define testcontainers here too?

I don't think so. Although it defines versions for things like TestNG and AssertJ, not every project using Airbase would require integration/e2e tests using containers.

kokosing:
Having testcontainers here does not require project using Aribase to use testcontainers. However, if they do use it, then they do not have to deal with testcontainers version anymore as it is already solved here.

nineinchnick:
I don't feel strongly about that, but most of my projects using airbase don't use testcontainers.

findepi (approved):

PR #323, opened 2022-08-15
Suppress Checkstyle within text-blocks

hashhar:
Some checks like RegexpMultiline are not AST aware and hence cannot
differentiate between content inside a text-block or outside of it. This
is a short term workaround which uses the """ text as a marker to let
Checkstyle suppress any violations between them.

The longer term fix is to migrate away from non-AST aware checks like
RegexpMultiline and RegexpSingleline to something else.

cc: @kokosing @martint 

Fixes https://github.com/trinodb/trino/issues/14623

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -3,6 +3,13 @@
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
 <module name="Checker">
+    <!-- TODO: Remove this supression once RegexpMultiline has been replaced with an AST aware check -->
+    <!-- This suprpresses Checkstyle between onccurences of """", i.e. text blocks -->
+    <module name="SuppressWithPlainTextCommentFilter">
+        <property name="offCommentFormat" value='=\s+"""'/>
+        <property name="onCommentFormat" value='^\s+.*""";'/>
+    </module>
+
     <module name="FileTabCharacter" />
     <module name="NewlineAtEndOfFile">
         <property name="lineSeparator" value="lf" />
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ Airbase 129
 
 * Checkstyle updates:
   - Require package name to match the source directory
+  - Suppress Checkstyle within text-blocks
 * Dependency updates:
   - joda-time 2.10.14 (from 2.10.13)
   - slf4j 1.7.36 (from 1.7.32)


hashhar:
I've been looking at some alternative checks to migrate the Regexp checks to. So far it seems EmptyLineSeparator can be configured to handle some cases but it doesn't cover everything.

If I don't find an existing check I'll try writing a new one that applies all vertical whitespace related rules we have today.

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -3,6 +3,13 @@
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
 <module name="Checker">
+    <!-- TODO: Remove this supression once RegexpMultiline has been replaced with an AST aware check -->
+    <!-- This suprpresses Checkstyle between onccurences of """", i.e. text blocks -->
+    <module name="SuppressWithPlainTextCommentFilter">
+        <property name="offCommentFormat" value='=\s+"""'/>
+        <property name="onCommentFormat" value='^\s+.*""";'/>
```
findinpath, line 10: Take into consideration the following use case:
- the text block is a parameter among other parameters to a method call (can be followed by `,` or `)`

hashhar: Hmm, that makes it slightly challenging. Since we need some way to differentiate opening and closing text block markers. One way to do this is to enforce all text-block content appears on a newline by changing start marker to `\s+"""$`.

Changing closing marker to `^\s+.*""".+` would mean it can still match opening marker if we don't enforce that text blocks start on a new line.

Any suggestions?
In meantime I'm looking at replacing RegexpMultiline.

findinpath: I guess your solution is based on the proposal

https://github.com/checkstyle/checkstyle/issues/11653#issuecomment-1180373074

What if we use something like:

```
<property name="onCommentFormat" value='^\s+.*"""(;|,|\))'/>
```

?

hashhar: That covers most of the cases we know of. Thanks. Hopefully this is a short term solution so we don't need to make it pretty.

findinpath: There are a few minor variations when dealing with `)`, it can be followed optionally by other `)` and eventually by `,` or `;`. Good thing is that AFAIK they are all on the same line.

```
<property name="onCommentFormat" value='^\s+.*"""(,|;|(\)+(,|;)))'/>
```

NOTE that I tested this so far only on regex101.com and not with the changes on your branch.

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -3,6 +3,13 @@
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
 <module name="Checker">
+    <!-- TODO: Remove this supression once RegexpMultiline has been replaced with an AST aware check -->
+    <!-- This suprpresses Checkstyle between onccurences of """", i.e. text blocks -->
```
findepi, line 7: nit: suprpresses -> suppresses

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
findepi: > * the text block is a parameter among other parameters to a method call (can be followed by `,` or `)`

i think this is important.
It's confusing to allow 

```
String x = """
   )
 """;
```

while not allow
```
assertQuery(
   """
    )
    """);
```

the text blocks are awesome for SQL queries and SQL queries are awesome to inline in tests, so this would be primary use-case

airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
```diff
@@ -3,6 +3,13 @@
         "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
         "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
 <module name="Checker">
+    <!-- TODO: Remove this supression once RegexpMultiline has been replaced with an AST aware check -->
```
findepi, line 6: What can be done here?

electrum:
@hashhar that would be great to write a new checker that can replace the regex checks. I've wanted to do that for a while. It's probably not hard to add a new Checkstyle module to Airbase, or possibly add it into the existing `airbase-policy` module.

hashhar:
I'd like to add it but unfortunately I'm not actively working on this at the moment. I'll probably get to it in the next week and it's free for someone else to take a shot at this until then.

PR #325, opened 2022-09-09
Update deps

nineinchnick:
Joda time and Checkstyle were recently updated and this PR bumps them again to the latest version available.

It also updates Jackson and JMH.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,11 @@ Airbase 129
   - modernizer 2.4.0 (from 2.3.0)
   - joda-time 2.10.14 (from 2.10.13)
   - slf4j 1.7.36 (from 1.7.32)
-  - checkstyle 10.3.2 (from 9.2)
+  - checkstyle 10.3.3 (from 9.2)
+  - Jackson 2.13.4 (from 2.13.3)
+  - JMH 1.35 (from 1.32)
+* Plugin updates:
+  - checkstyle 3.2.0 (from 3.1.2)
 
 Airbase 128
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -7,7 +7,7 @@ Airbase 129
   - modernizer 2.4.0 (from 2.3.0)
   - joda-time 2.10.14 (from 2.10.13)
   - slf4j 1.7.36 (from 1.7.32)
-  - checkstyle 10.3.2 (from 9.2)
+  - checkstyle 10.3.3 (from 9.2) and maven-checkstyle-plugin 3.2.0 (from 3.1.2)
```
electrum, line 10: The Maven plugin should be in a "Plugin updates" section

electrum, line 10: The "from" version is wrong, should be 10.3.2

CHANGES.md
```diff
@@ -5,7 +5,7 @@ Airbase 129
 * Dependency updates:
   - spotbugs-annotations 4.7.2 (from 4.3.0)
   - modernizer 2.4.0 (from 2.3.0)
-  - joda-time 2.10.14 (from 2.10.13)
+  - joda-time 2.11.1 (from 2.10.13)
```
electrum, line 8: The "from" version should be 2.10.14

CHANGES.md
nineinchnick: Done!

CHANGES.md
nineinchnick: This file lists changes vs the previous release and I updated joda-time recently in caa3e0847b47bce9a1a827fee0de83a5fdbe88da, which was not released yet. I tried explaining that in the PR description, but it probably could be made more clear.

CHANGES.md
nineinchnick: It's the same case as with joda-time, it was recently updated and was not released yet.

nineinchnick (commented):
I fixed commit messages too.

PR #329, opened 2022-11-16
Use ISO-8859-1 when filtering properties files

findepi:
Properties files are standardized to be encoded with ISO-8859-1. maven-resources-plug now allows setting a specific encoding for properties files and even generates a build-time message when this configuration is not used.

This can be e.g. observed in a Trino build

    [INFO] --- maven-resources-plugin:3.2.0:resources (default-resources) @ trino-testing-services ---
    [INFO] Using 'UTF-8' encoding to copy filtered resources.
    [INFO] Using 'UTF-8' encoding to copy filtered properties files.
    [INFO] Copying 1 resource
    [INFO] Copying 1 resource
    [INFO] The encoding used to copy filtered properties files have not
    been set. This means that the same encoding will be used to copy
    filtered properties files as when copying other filtered resources. This
    might not be what you want! Run your build with --debug to see which
    files might be affected. Read more at
    https://maven.apache.org/plugins/maven-resources-plugin/examples/filtering-properties-files.html

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 144
+
+* Use standard properties encoding, ISO-8859-1, when filtering properties file.
+
 Airbase 143
 
 * Relax restricted imports for Jersey
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -470,12 +470,14 @@
                     </configuration>
                 </plugin>
 
-                <!-- Resource plugins should always use UTF-8 -->
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-resources-plugin</artifactId>
                     <version>3.3.1</version>
                     <configuration>
+                        <!-- Properties files are ISO-8859-1 by default -->
+                        <propertiesEncoding>ISO-8859-1</propertiesEncoding>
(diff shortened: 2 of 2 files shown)


martint:
I don't think we want this. We should standardize on UTF-8 instead. 

Since Java 9, resource bundles for internationalization (which are also properties files) are expected to be UTF-8 by default (see https://openjdk.org/jeps/226). Incidentally, this change would make it hard to use property resource bundles in an airlift-based project.

martint:
IntelliJ docs say:

> [...] Alternatively, you can define the default encoding for properties files on the project level and use a different API that can read properties files in the encoding you have defined.

https://www.jetbrains.com/help/idea/properties-files.html#encoding

I wonder if there's a way to make it adopt that setting from observing something in the pom, similar to how it handles the JDK version to use for the project.

findepi:
in theory, this should already do it

https://github.com/trinodb/trino/blob/70d02427402818528cf8e76f0f760d54ed64df1c/.editorconfig#L4-L5

findepi:
rebased and resolved conflict in CHANGES.

note that this addresses build-time warning from maven.

findepi:
maven prints this warning-like message at INFO, so i will just ignore it

PR #346, opened 2023-04-06
Pin Jackson module versions using the official BOM

ksobolew:
This module guarantees that we declare all the possible Jackson modules and we don't have to enumerate them. Previously we missed some important ones, like `jackson-dataformat-xml`.


electrum (approved):

electrum:
Nice!

PR #351, opened 2023-04-13
Update dependencies for Java 20

electrum:


airbase/pom.xml
```diff
@@ -416,6 +419,7 @@
                         <showDeprecation>true</showDeprecation>
                         <showWarnings>true</showWarnings>
                         <parameters>true</parameters>
+                        <enablePreview>${air.compiler.enable-preview}</enablePreview>
```
dain, line 422: Is it possible to automatically add the surefire argline when this is enabled?

dain (approved):

airbase/pom.xml
electrum: Let me see if we can do that with a profile

dain (approved):

PR #355, opened 2023-05-10
Add shaded testcontainers to illegal imports

wendigo:

PR #357, opened 2023-05-11
Replace checkstyle illegal import rules with RestrictImports

wendigo:
Alternative approach to https://github.com/airlift/airbase/pull/355 as suggested by @electrum 

For reference: https://github.com/skuzzle/restrict-imports-enforcer-rule

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+Airbase 139
+
+* Replace checkstyle illegal imports with `RestrictImports` enforcer rules
+
 Airbase 138
 
 * Dependency updates:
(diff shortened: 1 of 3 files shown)


wendigo:
cc @electrum @hashhar @kokosing @findepi

hashhar:
errorprone profile isn't enabled by default locally so devs won't catch until these reach CI - just something to consider

wendigo:
@hashhar enforcer is enabled on a default profile. I've checked that PR with the minimal set of flags

electrum (approved):

hashhar:
> @hashhar enforcer is enabled on a default profile. I've checked that PR with the minimal set of flags

Ah, I was confused between errorprone and enforcer plugins. 😄

PR #358, opened 2023-05-25
Fix running timezone-sensitive tests in IntelliJ IDEA

ksobolew:
Some time ago, some properties were moved to `argLine` because Surefire incorrectly passes the `user.language` and `user.region` system properties to the tests. At that time, the `user.timezone` also got along for the ride, but it turns out, that that latter property *is* passed correctly. So while it works for Maven and Surefire, moving it into `argLine` causes IntelliJ IDEA to not pick it up, which in turn means that some tests are not working properly when executed inside the IDE. We can move it safely back, as it improves quality of life.

This reverts parts of commit d4630aa6547a64c7112f2bf133138d7d83b6d1a8.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+Airbase 140
+
+* Pass `user.timezone` in tests as a system property, not as a JVM argument
+  - It restores the ability to run some timezone-sensitive tests
+    from within IntelliJ IDEA
+
 Airbase 139
 
 * Replace checkstyle illegal imports with `RestrictImports` enforcer rules
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -498,6 +498,7 @@
                     <configuration>
                         <systemPropertyVariables>
                             <sun.jnu.encoding>${project.build.sourceEncoding}</sun.jnu.encoding>
+                            <user.timezone>${air.test.timezone}</user.timezone>
                             <java.awt.headless>true</java.awt.headless>
                             <java.util.logging.SimpleFormatter.format>%1$tY-%1$tm-%1$tdT%1$tH:%1$tM:%1$tS.%1$tL%1$tz %4$s %5$s%6$s%n</java.util.logging.SimpleFormatter.format>
                         </systemPropertyVariables>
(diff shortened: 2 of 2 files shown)


airbase/pom.xml
```diff
@@ -512,7 +513,6 @@
 
                             <!-- Pass the following three directly to the JVM,
```
findepi, line 514: "three" ?

ksobolew: Hah... Proof that nobody looks at comments (except a reviewer, sometimes) :) Will change

ksobolew: Done

findepi:
thanks

update CHANGES too

airbase/pom.xml
ksobolew: > if you feel no-one is going to read it, why would we keep it at all?

I was just being snarky, sorry. After X years in this business my brain apparently got trained to filter out comments and I just don't see them anymore, unless I make an effort.

ksobolew:
> update CHANGES too

Done

ksobolew:
> > update CHANGES too
> 
> Done

...under the correct version now

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+Airbase 140
+
+* Pass `user.timezone` in tests as a system property, not as a JMV argument
```
findepi, line 3: what does JMV stand for?

also mention that this supposedly restores IntelliJ ability to set timezone correctly when running tests from within that IDE.

findepi (approved):

martint:
Can you update the commit message to say “fix xxxx”? You can mention that it’s a partial revert of something else in the commit message, but that’s unnecessary for the title of the commit.

ksobolew:
> Can you update the commit message to say “fix xxxx”? You can mention that it’s a partial revert of something else in the commit message, but that’s unnecessary for the title of the commit.

Sure; done

findepi:
the new commit title is

> [Fix passing timezone information to Surefire](https://github.com/airlift/airbase/pull/358/commits/6a9a23caf630610ae844f88d318972f91fdba49b)

which isn't exactly 100% correct.

we were passing timezone information to surefire correctly
the fix is for Intellij, afaiu

ksobolew:
> we were passing timezone information to surefire correctly
> the fix is for Intellij, afaiu

You're right, I changed the title and reworded the commit message a bit too.

PR #360, opened 2023-06-01
Update extra-enforcer-rules to 1.7.0

wendigo:
This version recognized JDK 21 classes correctly (previous one fails on JDK 21 with target 21)


wendigo:
cc @martint: please review, merge and release so I can update airbase in Trino. This is the last plugin that wasn't working with JDK 21. error-prone needs updating but it's managed in the Trino pom.

PR #361, opened 2023-06-06
Use jakarta for ee dependencies

wendigo:
This defines Jakarta dependencies to use by downstream projects:

- rs.ws-api
- xml.bind-api
- servlet-api
- annotation-api


airbase/pom.xml
```diff
@@ -1130,21 +1132,29 @@
...
+                <artifactId>jakarta.inject-api</artifactId>
+                <version>${dep.jakarta-inject.version}</version>
             </dependency>
+
+            <!-- servlet -->
             <dependency>
-                <groupId>javax.validation</groupId>
-                <artifactId>validation-api</artifactId>
-                <version>${dep.javax-validation.version}</version>
+                <groupId>jakarta.servlet</groupId>
```
electrum, line 1142: Per the Jetty team, we want to use `jetty-jakarta-servlet-api`. When 6.0 is out, we can switch to the official `jakarta.servlet` dependency.

airbase/pom.xml
```diff
@@ -339,6 +339,8 @@
                                     <!-- use the newer Jetty version of the servlet API -->
                                     <exclude>javax.servlet:javax.servlet-api</exclude>
                                     <exclude>org.eclipse.jetty.orbit:javax.servlet</exclude>
+                                    <exclude>org.eclipse.jetty.toolchain:jetty-servlet-api</exclude>
```
electrum, line 342: We shouldn't exclude this one, as it's a separate artifact and Java package which might be needed by dependencies. It doesn't conflict with the new ones.

airbase/pom.xml
```diff
@@ -1130,21 +1132,29 @@
 
             <!-- Java APIs -->
             <dependency>
```
electrum, line 1134: Let's keep the old one around, to ease migration

airbase/pom.xml
```diff
@@ -1130,21 +1132,29 @@
...
+            <!-- xml bind -->
+            <dependency>
+                <groupId>jakarta.xml.bind</groupId>
+                <artifactId>jakarta.xml.bind-api</artifactId>
+                <version>4.0.0</version>
             </dependency>
 
-            <!-- servlet -->
             <dependency>
-                <groupId>org.eclipse.jetty.toolchain</groupId>
```
electrum, line 1145: Same, keep all the existing dependencies

PR #369, opened 2023-06-20
Update maven-shade-plugin to 3.5.0

ksobolew:
This release, among other things, natively depends on ASM 9.5 with support for new opcodes in Java 21, but we'll keep the ASM version override in case we need a blanket ASM upgrade in the future.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Airbase 143
   - jackson 2.15.2 (from 2.15.0)
 * Plugin updates:
   - maven-gpg-plugin 3.1.0 (from 1.4.0)
+  - maven-shade-plugin 3.5.0 (from 3.4.1)
 
 Airbase 142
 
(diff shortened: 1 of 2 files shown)


ksobolew:
cc: @wendigo

ksobolew:
Though one thing I notice, is that it seems to create `dependency-reduced-pom.xml` in the module root directory. Not sure if I should treat it as a bug? Shouldn't it be in `target/` instead?

wendigo:
Cc @kokosing

electrum:
This seems to be the documented behavior. We typically use the following:
```xml
<dependencyReducedPomLocation>${project.build.directory}/pom.xml</dependencyReducedPomLocation>
```

ksobolew:
> This seems to be the documented behavior. We typically use the following:
> 
> ```
> <dependencyReducedPomLocation>${project.build.directory}/pom.xml</dependencyReducedPomLocation>
> ```

Yeah, but somehow this file wasn't actually created before, at least for me. Maybe it was broken in earlier versions 🤷

wendigo:
Because previously reduced pom was not created. Now it is and it breaks usage of modules that are both executable and libraries like product test launcher in Trino

wendigo:
It's already configured in Trino where needed so I guess that we are fine

PR #374, opened 2023-07-14
Fix error-prone version

wendigo:


wendigo:
cc @electrum @martint

PR #382, opened 2023-09-30
Downgrade logback to 1.4.8

wendigo:
There is a regression in logback 1.4.9 which causes Trino tests to fail due to timeouts. The root cause is still to be investigated.

Related: https://github.com/trinodb/trino/pull/19194
Trino logback update test: https://github.com/trinodb/trino/pull/19207


wendigo:
cc @electrum

PR #384, opened 2023-10-13
Update dependencies & plugins

wendigo:
Compilation tested against both Trino and Airlift

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Airbase 148
   - errorprone 2.22.0 (from 2.21.1)
   - Guava 32.1.3-jre (from 32.1.2-jre)
   - spotbugs-annotations 4.8.0 (from 4.7.3)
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -5,6 +5,9 @@ Airbase 148
 * Dependency updates:
   - Jackson 2.15.3 (from 2.15.2)
 
+* Plugin updates:
+  - errorprone 2.22.0 (from 2.21.1)
```
electrum, line 13: This is actually a dependency update. We don't (yet) have an Error Prone plugin configuration in Airbase.

electrum (approved):

CHANGES.md
wendigo: I'll correct changelog

wendigo:
@electrum fixed, I've also added new jacoco plugin which supports JDK 21

wendigo:
I've added 3 more plugin updates

PR #389, opened 2024-01-10
Forbid code that would reformat poorly

findepi:
Current checkstyle rules somehow allow things like

    void run() { throw new UnsupportedOperationException(); }

    Runnable r = () -> { throw new RuntimeException(); }

    Runnable asRunnable()
    { a_field++; }

    void update()
    { throw new RuntimeException(); }

however, when such code is then auto-formatted by IntelliJ with Airbase codestyle, some whitespace is removed, resulting in a code that violates checkstyle rules. This state is undesired: sometimes a person working on a class file may need to undo auto-formatting so pass checkstyle checks.

The new rules aim to reduce such situation by recognizing some common would-format-incorrectly patterns.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1162,4 +1162,3 @@ Airbase 2
 Airbase 1
 
 * Initial release
-
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 Airbase 149
 * Remove bval dependency
+* Checkstyle updates:
+  - Disallow single-line non-empty lambda and method body code blocks.
 * Plugin updates:
   - git-commit-id-maven-plugin 7.0.0 (from 6.0.0)
   - spotbugs-maven-plugins 4.8.1.0 (from 4.7.3.6)
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -256,6 +256,22 @@
             <property name="ignoreComments" value="true" />
         </module>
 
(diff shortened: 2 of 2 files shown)


martint (changes requested):
This is a readable and concise way to format the code, so we shouldn't prevent it:

```java
Runnable r = () -> { throw new RuntimeException(); }
```

I'm not experiencing the issues with whitespace re-formatting that you mention, so I would suggest diagnosing that first.

findepi:
I experience this in all recent Intellij versions.
How do i go about diagnosing that?

martint:
Look at the codestyle options and try to see which one is causing that? Maybe they introduced a new option that's defaulting to some other setting...

martint:
FWIW, I'm running IntelliJ 2023.3.2 and I don't experience the problem there.

findepi:
i already tried that, but didn't find an option for this.
i am still at 2023.2.x (because 2023.3.2 did not appear to work stably for me), but will try 2023.3.2 (ultimate) too

findepi:
switching to 2023.3.2 and importing https://github.com/airlift/codestyle/blob/master/IntelliJIdea2019/Airlift.xml anew did not make intellij retain code as is upon auto-formatting

```java
Runnable r = () -> { throw new RuntimeException(); }
```

is there any additional configuration i should set to make it work?

martint:
Do you have Spaces->Within->Code Braces turned on?

findepi:
> Do you have Spaces->Within->Code Braces turned on?

this indeed helps!

https://github.com/airlift/codestyle/pull/15

PR #395, opened 2024-02-21
Updates, updates, updates 

wendigo:


wendigo:
PTAL @martint @electrum

wendigo:
Tested locally with Trino. Builds without any changes.

wendigo:
@martint @electrum ptal.

PR #396, opened 2024-03-07
Downgrade modernizer to 2.7.0

wendigo:
2.8.0 requires maven 3.9.6 and doesn't work with mvnd in the process.


wendigo:
@electrum ptal

electrum:
Is there an issue for mvnd?

wendigo:
@electrum The issue is that mvnd is an old release

PR #402, opened 2024-03-17
Update dependencies

wendigo:


wendigo:
PTAL @electrum. This makes it possible to build with target language level 22

findepi (approved):

wendigo:
@electrum @martint PTAL. I'd like to update it in Trino and test with JDK 22-ea

wendigo:
I've tested these changes with Airlift and Trino and seems fine (code needs some fixes due to deprecated guava's `Throwables` methods but these are minor)

PR #404, opened 2024-04-04
Fix Intellij's inference of preview language level

findepi:


findepi:
cc @wendigo

findepi:
@wendigo was https://github.com/airlift/airbase/pull/405#issuecomment-2037823722 supposed to mean **this** PR does not work?

findepi:
indeed the property generated with build-helper-maven-plugin is not picked by intellij, otherwise it would work

PR #405, opened 2024-04-04
Fix javadoc plugin configuration when JDK preview features used

findepi:


findepi:
cc @electrum @wendigo

findepi:
relates to https://github.com/airlift/airbase/pull/404

wendigo:
This still doesn't work in IntelliJ. I think that this is due to the fact that `air.compiler.enable-preview` is evaluated to produce `surefireEnablePreview` in the `build-helper-maven-plugin` which is not something that IntelliJ understands and support.

wendigo:
I think that instead of `air.compiler.enable-preview` we should introduce `air.compiler.extra-args`

wendigo:
@findepi this would work

findepi:
@electrum would you like to take a look?

electrum (approved):

PR #403, opened 2024-03-22
Improved JDK 22-23 compatibility, reproducible builds support

wendigo:
PMD 7.0.0 is finally compatible with JDK 21

PR #406, opened 2024-04-10
Support reproducible maven builds

wendigo:
https://maven.apache.org/guides/mini/guide-reproducible-builds.html#faq


wendigo:
@electrum @martint PTAL

wendigo:
Migrated to https://github.com/airlift/airbase/pull/403

PR #411, opened 2024-05-29
Disallow class-level warning suppression 

findepi:
`@SuppressWarnings` should be applied to narrowest scope possible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 Airbase 157
 * Checkstyle updates:
+ - Disallow class-level warning suppression
  - Require empty line before record definition
 
 Airbase 156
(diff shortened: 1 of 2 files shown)


wendigo:
Did you try that on the Trino?

findepi:
> Did you try that on the Trino?

https://github.com/trinodb/trino/pull/22193

wendigo (approved):
LGTM but you'll need to apply it in Trino and Airlift :)

wendigo:
Sometimes these suppressions are actually needed. In Airlift: 
```
@SuppressWarnings("checkstyle:TypeName")
    public static class $$RecordedConfigData<T>
```

findepi:
I am not convinced this is how we would name a class in 2024.

electrum:
This seems heavy-handed

findepi:
I can't disagree.

PR #413, opened 2024-06-07
fix Reproducible Builds issue

hboutemy:
see https://github.com/jvm-repo-rebuild/reproducible-central/tree/master/content/io/airlift/airbase

### DIFF
--- a/airbase-policy/pom.xml
+++ b/airbase-policy/pom.xml
@@ -14,5 +14,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


wendigo (approved):

airbase-policy/pom.xml
```diff
@@ -14,5 +14,6 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.build.outputTimestamp>2024-06-05T18:08:46Z</project.build.outputTimestamp>
```
findepi, line 17: can `project.build.outputTimestamp` be injected only for releases?

wendigo: Why are you asking?

PR #414, opened 2024-06-17
Require maven 3.9.8 to use airbase

wendigo:
3.9.8 is version used by mvnd 1.0.0 so let's make it a minimum required version which also means that we will support mvnd 1.0.0+ out of the box.


airbase/pom.xml
```diff
@@ -148,7 +148,7 @@
         <air.check.fail-checkstyle>${air.check.fail-basic}</air.check.fail-checkstyle>
 
         <!-- minimum maven version to build -->
-        <air.maven.version>3.6.3</air.maven.version>
```
findepi, line 151: What's the benefit of not allowing to run the build with 3.6

airbase/pom.xml
```diff
@@ -148,7 +148,7 @@
...
+        <air.maven.version>3.9.8</air.maven.version>
```
findepi, line 151: How old is 3.9.8?

wendigo: It's not about how old it is. I'd like to see mvnd supported across the board and 1.0.0 (reporting as 3.9.8) is the first version I actually trust to deliver a value

findepi: Age matters. if 3.9.8 is relatively recent and for example we uncover some edge case bug in 3.9.8 and there is no other version we can use, we actively prevented ourselves from using an older version as a workaround

wendigo: This can be overridden if needed with the property.

mosabua (approved):
Maven releases have been getting more and more stable and there are a lot of good refactors between 3.6.3 and 3.9.8. 

Definitely agree with this upgrade.

wendigo:
That's my thinking. I've invested a lot of time into getting our maven infrastructure modern, simple, performant, and reproducible. Maven 3.9.8 along with the mvnd 1.0.0 (built on top of the 3.9.8) marks the culmination of this effort. I'd like to enforce the usage of these versions across our projects to have consistent behavior and also to claim that mvnd is supported and recommended for local development.

wendigo:
@electrum @dain @martint please chime in

wendigo:
The same discussion happens in the Quarkus community: https://groups.google.com/g/quarkus-dev/c/XU8i2tCEVPA

findepi:
> The same discussion happens in the Quarkus community: https://groups.google.com/g/quarkus-dev/c/XU8i2tCEVPA

thanks for the link, this is really interesting!

I love how this thread starts with a clearly defined rationale:
_"We are not usually being too aggressive for our Maven requirements but... in this case we have a very good reason: Apache Maven 3.9.5- doesn't support Maven plugins compiled with Java 17."_

wendigo:
@findepi I wasn't aware of that problem and I think that we only use plugins compiled with JDK8 (which is required for Maven 3.x)

electrum (approved):

slachiewicz:
While updating Maven Wrapper to use the latest version of Maven is a good thing - requirement for all developers to use the latest version may be too aggressive.
Saying that You require Maven 3.9.x should be good enough.

In theory the Maven community supports Maven 3.6.3+ and we migrate all our base plugins to that version.

mosabua:
@slachiewicz sure... but if we want to require 3.9.8 for airbase and any projects inheriting it would make our support for all those projects easier .. less testing and verification work. And also with wrapper config and such its completely transparent and easy to use whatever Maven version. Our move to require 3.9.8 is also explicitly to help the Maven project with testing and moving towards 4.0.0. We want to ensure that the latest and greatest Maven version is the best .. and dont really care about old stuff.

electrum:
I support requiring the latest version. We can pretend that we support 3.6.3 but in reality we only test against the latest. Unless someone has a specific objection to upgrading (vs theoretical problems) then I’m good with this.

mosabua:
Let's ship this @electrum and @wendigo !

PR #420, opened 2024-08-05
Add jreleaser override for airbase-policy

wendigo:


wendigo:
First release attempt failed because `airbase-policy` doesn't have javadoc or sources attached

electrum (approved):

PR #816, opened 2025-07-07
Use legacy altDeploymentRepository format for better compatibility

wendigo:
# Airbase contribution check list

 - [x] Git commit messages follow https://cbea.ms/git-commit/.
 - [x] All automated tests are passing.
 - [x] Pull request was categorized using one of the existing labels.

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -1739,6 +1739,26 @@
         </profile>
 
         <profile>
+            <id>javadoc-ge-jdk-23</id>
+            <activation>
+                <jdk>[23,)</jdk>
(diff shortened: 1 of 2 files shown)


cstamas:
For context: Maven historically supported `id::layout::url` form, but since Maven 3 there is only "default" layout, so m-deploy-p was modified to support shorter `id::url` format, but https://github.com/takari/takari-lifecycle/issues/214

wendigo:
That works but fails due to sonatype central limitations :) (good to merge anyway)

electrum (approved):

PR #1069, opened 2026-01-29
Allow flexible constructor bodies

findepi:
Java 25 adds ability to place statements within constructor before call to `super()`. Checkstyle does not support this yet, so it's configured not to fail the build on these.

### DIFF
--- a/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
+++ b/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml
@@ -103,6 +103,18 @@
     <module name="SuppressWarningsFilter" />
 
     <module name="TreeWalker">
+        <!--
+          Suppress Checkstyle Java code parse errors.
+          TODO (https://github.com/checkstyle/checkstyle/issues/17052) remove suppression when checkstyle supports flexible constructor bodies (JEP 513)
+          -->
+        <property name="skipFileOnJavaParseException" value="true" />
+        <!--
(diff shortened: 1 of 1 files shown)


wendigo (approved):

wendigo (approved):

findepi:
> This means we don't get any style checks for these files. 

Correct

> Is it so important to use this feature before checkstyle supports it?

If checkstyle adds support for this soon, it doesn't matter.
If checkstyle doesn't add support for this soon, do we want to be held back?

Checkstyle is advisory only anyway.  The fact the code passes checkstyle doesn't unfortunately prove it won't change on Intellij's reformat, which is quite useful property to have.

findepi:
@martint wdyt?

martint:
We should allow them, yes. Now that we have the airlift formatter, do we even need checkstyle?

electrum:
This is supported in checkstyle now: https://github.com/checkstyle/checkstyle/pull/19364

electrum:
@martint see the [current checkstyle rules](https://github.com/airlift/airbase/blob/master/airbase-policy/src/main/resources/checkstyle/airbase-checks.xml)

None of those rules are formatting rules. Airstyle is a pure code formatter, not a linter. It can format the code, or it can tell you that the code does not match what the formatter would output.

The checkstyle rules cover things like variable naming. It’s not appropriate for a code formatter to rename your variables.

findepi:
> This is supported in checkstyle now: [checkstyle/checkstyle#19364](https://github.com/checkstyle/checkstyle/pull/19364)

PR #1070, opened 2026-01-30
Disallow use of awaitility library

findepi:
The awaitility library mutates JVM state and affects isolation between JUnit tests. Its use is discouraged.  For example in Trino it caused `TestNodeStateManager` test to become flaky by intercepting exceptions from threads spawned by a different test. The two tests did not share any common executor or anything.  Likely the reason for this is the fact the library installs a global exception handler.

While this awaitility's behavior is configurable, it's also the default. It is also not clear whether this is the only problem. Discourage use of the library.

- follows https://github.com/trinodb/trino/issues/25705#issuecomment-3823539429

### DIFF
--- a/airbase/pom.xml
+++ b/airbase/pom.xml
@@ -858,6 +858,14 @@
                                     <bannedImport>org.assertj.core.api.AssertionsForInterfaceTypes.*</bannedImport>
                                 </bannedImports>
                             </RestrictImports>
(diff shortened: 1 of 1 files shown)


findepi:
Switched to `<RestrictImports>` and added a link to explaination

- https://github.com/awaitility/awaitility/issues/120#issuecomment-420977109

PR #1220, opened 2026-05-26
Remove Checkstyle rules covered by Airstyle

electrum:
Removes duplicate Checkstyle formatting enforcement for rules now handled by Airstyle.

Checkstyle remains focused on non-formatting policy and correctness checks.


wendigo (approved):

wendigo:
Cool. These regex rules were pretty slow. This should make it much faster

PR #1271, opened 2026-06-24
Restore `*` import checkstyle check

findepi:
This is not covered by airstyle yet.


wendigo (approved):

findepi:
We can also disallow this with error-prone's `WildcardImport`, if that's preferred.

PR #1304, opened 2026-07-20
Add OpenTelemetry API incubator

electrum:
Adds the OpenTelemetry incubator API dependency so downstream projects can use its incubating APIs.


wendigo (approved):

electrum:
Closing as redundant: the imported OpenTelemetry alpha BOM already manages opentelemetry-api-incubator at the same version.

PR #1319, opened 2026-08-06
Disable cooldown while updating airbase packages

wendigo:
# Airbase contribution check list

 - [x] Git commit messages follow https://cbea.ms/git-commit/.
 - [x] All automated tests are passing.
 - [x] Pull request was categorized using one of the existing labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    cooldown:
+      default-days: 0
     labels:
       - "dependency"
     open-pull-requests-limit: 5


Randgalt (approved):

.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    cooldown:
+      default-days: 0
```
findepi, line 8: isn't this less safe?

findepi: @electrum wdyt about security posture?

wendigo: I review every dependency that I update

